PR #59, opened 2025-01-13
Add FCS resources

agandrabur:
# Add Falcon Cloud Security Resources and Data Sources

This PR adds support for managing AWS accounts in Falcon Cloud Security
### `crowdstrike_cloud_aws_account`
Manages AWS accounts in Falcon cloud security with features including:
- Account registration with organization support
- Real-time visibility configuration
- Sensor management
- DSPM (Data Security Posture Management) settings
- Identity Protection

### `crowdstrike_cloud_aws_accounts`
Retrieves information about AWS accounts in Falcon cloud security including:
- Account details
- Organization settings

Can filter by AWS account ID or AWS organization ID


internal/fcs/cspm_aws_account.go
```diff
@@ -0,0 +1,700 @@
...
+
+// Schema defines the schema for the resource.
+func (r *cspmAWSAccountResource) Schema(
+	_ context.Context,
+	_ resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: fmt.Sprintf(
+			"CSPM AWS Account --- This resource allows management of a CSPM Account. A FileVantage policy is a collection of file integrity rules and rule groups that you can apply to host groups.\n\n%s",
```
ffalor, line 88: It looks like this may include the description pasted from another resource

internal/fcs/fcs_aws_account.go
```diff
@@ -0,0 +1,607 @@
...
+
+// Schema defines the schema for the resource.
+func (r *cloudAWSAccountResource) Schema(
+	_ context.Context,
+	_ resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: fmt.Sprintf(
+			"FCS AWS Account --- This resource allows management of a CSPM Account. A FileVantage policy is a collection of file integrity rules and rule groups that you can apply to host groups.\n\n%s",
```
ffalor, line 75: It looks like this may include the description pasted from another resource

ffalor (changes requested):

ffalor:
we will need an example, they are used in documentation

internal/fcs/cloud_aws_account_data_source.go
```diff
@@ -0,0 +1,438 @@
...
+	EventbusName           types.String                `tfsdk:"eventbus_name"`
+	EventbusArn            types.String                `tfsdk:"eventbus_arn"`
+	CloudTrailRegion       types.String                `tfsdk:"cloudtrail_region"`
+	CloudTrailBucketName   types.String                `tfsdk:"cloudtrail_bucket_name"`
+	DspmRoleArn            types.String                `tfsdk:"dspm_role_arn"`
+	AssetInventory         *asssetInventoryFeature     `tfsdk:"asset_inventory"`
+	RealtimeVisibility     *reatltimeVisibilityFeature `tfsdk:"realtime_visibility"`
+	IDP                    *idpFeature                 `tfsdk:"idp"`
+	SensorManagement       *sensorManagementFeature    `tfsdk:"sensor_management"`
+	DSPM                   *dspmFeature                `tfsdk:"dspm"`
```
ffalor, line 72: are these structs because they will have more fields in the future?

internal/fcs/cloud_aws_account_data_source.go
ffalor, line 1: what if someone wants to get all accounts associated with a cid? Does the api support this (not providing an org id or account id filter)? If it does the resource should too.

internal/fcs/cloud_aws_account_data_source.go
```diff
@@ -0,0 +1,438 @@
...
+				fmt.Sprintf("Failed to get Cloud AWS accounts: %s", *error.Message),
+			)
+		}
+		return nil, diags
+	}
+	return res.Payload.Resources, diags
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (d *cloudAwsAccountsDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
```
ffalor, line 301: If we get a 403, lets respond with the needed scope also 

image

internal/fcs/cloud_aws_account_data_source.go
```diff
@@ -0,0 +1,438 @@
...
+// Schema defines the schema for the data source.
+func (d *cloudAwsAccountsDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Fetches Cloud AWS accounts by account_id or organization_id",
+		MarkdownDescription: fmt.Sprintf(
+			"Cloud AWS Accounts --- This data source provides information about AWS accounts in Falcon.\n\n%s",
+			scopes.GenerateScopeDescription(cloudSecurityScopes),
+		),
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
ffalor, line 101: ID isn't actually needed so this can be dropped all together. The other data resource we have has an ID, but that should have also been dropped.

internal/fcs/cloud_aws_account.go
```diff
@@ -0,0 +1,1217 @@
...
+	RoleName types.String `tfsdk:"role_name"`
+}
+
+type cloudAWSAccountModel struct {
+	AccountID              types.String               `tfsdk:"account_id"`
+	OrganizationID         types.String               `tfsdk:"organization_id"`
+	TargetOUs              types.List                 `tfsdk:"target_ous"`
+	IsOrgManagementAccount types.Bool                 `tfsdk:"is_organization_management_account"`
+	AccountType            types.String               `tfsdk:"account_type"`
+	DeploymentMethod       types.String               `tfsdk:"deployment_method"`
```
ffalor, line 68: Is this something that is not going away anytime soon? If it's something that may no longer exist with the new apis I say we don't implement it. If it's staying then its fine to keep.

internal/fcs/cloud_aws_account.go
```diff
@@ -0,0 +1,1217 @@
...
+				},
+			},
+			"account_type": schema.StringAttribute{
+				Optional:    true,
+				Default:     stringdefault.StaticString("commercial"),
+				Computed:    true,
+				Description: "The type of account. Not needed for non-govcloud environment",
+				Validators: []validator.String{
+					stringvalidator.OneOf("commercial", "gov"),
+				},
```
ffalor, line 181: This is required if the falcon cloud is us-gov correct? If so then in `ValidateConfig` we can check what cloud is set. 

We will need to update the provider to also return `falcon.ApiConfig` and then we can use `apiConfig.Cloud` & if it's one of the gov clouds we can require `account_type`

internal/fcs/cloud_aws_account.go
ffalor, line 1: Anywhere that can 403 should return scope information

internal/fcs/cloud_aws_account_data_source.go
agandrabur: maybe, but not for all features

I will change the fields to booleans

internal/fcs/cloud_aws_account_data_source.go
agandrabur: Yes, the API supports it.
Providing an `account_id` and `organization_id` will however do an OR operation. 

Will update description to reflect that

internal/fcs/cloud_aws_account_data_source.go
agandrabur: removing

internal/fcs/cloud_aws_account.go
agandrabur: this is actually a new field, we want to keep it

internal/fcs/cloud_aws_account.go
agandrabur: backend will set it correctly and doesn't allow `gov` in commercial clouds

ffalor (approved):

PR #80, opened 2025-03-19
poll for eventbus arn when creating

ffalor:
- adds polling logic to ensure eventbus_arn is set when creating
- uses stateforunknown for computed variables to prevent inconsistent plans
- sets organization id on read to prevent import errors

### DIFF
--- a/internal/fcs/cloud_aws_account.go
+++ b/internal/fcs/cloud_aws_account.go
@@ -1352,6 +1352,28 @@ func (r *cloudAWSAccountResource) ImportState(
 ) {
 	// Retrieve import ID and save to id attribute
 	resource.ImportStatePassthroughID(ctx, path.Root("account_id"), req, resp)
+	account, diags := r.getCSPMAccount(ctx, req.ID)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if account.OrganizationID != "" {
+		resp.Diagnostics.Append(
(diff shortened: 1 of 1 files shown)


internal/fcs/cloud_aws_account.go
```diff
@@ -729,7 +745,7 @@ func (r *cloudAWSAccountResource) Read(
 	}
 
 	state.AccountID = types.StringValue(cspmAccount.AccountID)
-	state.OrganizationID = oldState.OrganizationID
+	state.OrganizationID = types.StringValue(cspmAccount.OrganizationID)
```
agandrabur, line 748: this will not work

If we register a child account first and then onboard the org management account, then this field will be populated. On the next `terraform apply` it will detect changes and will force re-create/replace. On next healthchecks the same will happen, on and on.

This is why I used the old state. We set it in terraform and then later it can be part of an org and not trigger force replace

internal/fcs/cloud_aws_account.go
```diff
@@ -1351,3 +1367,42 @@ func (r *cloudAWSAccountResource) ValidateConfig(
...
+		select {
+		case <-ticker.C:
+			account, diag := r.getCSPMAccount(pollCtx, config.AccountID.ValueString())
+			if diag.HasError() {
+				return diag
+			}
+
+			if account.AwsEventbusArn != "" {
+				config.EventbusArn = types.StringValue(account.AwsEventbusArn)
+				config.EventbusName = types.StringValue(account.EventbusName)
```
agandrabur, line 1394: and `config.CloudTrailBucketName`

agandrabur (changes requested):

PR #96, opened 2025-05-21
fix_static_host_group_doco: Update host_group.md to fix the example f…

manasmbellani:
…or static host groups

### DIFF
--- a/docs/resources/host_group.md
+++ b/docs/resources/host_group.md
@@ -50,7 +50,7 @@ resource "crowdstrike_host_group" "static" {
   description = "Made with terraform"
   name        = "Static Host Group rename"
   type        = "static"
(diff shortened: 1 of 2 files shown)


ffalor:
@manasmbellani thank you very much for catching this and letting us know. 

The example in that doc comes from [this file ](https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/main/examples/resources/crowdstrike_host_group/resource.tf) so your changes will need to be made there and than we can generate the docs using go generate.

manasmbellani:
Hello @ffalor , I have updated the code. Thanks.

ffalor (approved):

ffalor:
Thank you for the fix!

PR #118, opened 2025-08-01
Add crowdstrike_content_category_versions data source

ffalor:
Implements a new data source to retrieve available content versions for different categories (sensor operations, system critical, vulnerability management, rapid response) that can be used for pinning in content update policies.


ffalor:
closes #119

PR #147, opened 2025-09-10
feat: Add cloud posture custom resource and cloud rules data source

gpontejos:


internal/cloud_posture/custom_rule_resource_test.go
```diff
@@ -0,0 +1,359 @@
...
+		}, "")
+		remediationInfo := strings.Join([]string{
+			`"` + strings.Join(config.ruleBaseConfig.remediationInfo[i], `","`) + `"`,
+		}, "")
+		attackTypes := strings.Join([]string{
+			`"` + strings.Join(config.ruleBaseConfig.attackTypes[i], `","`) + `"`,
+		}, "")
+		resourceStep := resource.TestStep{
+			Config: fmt.Sprintf(`
+resource "crowdstrike_cloud_posture_custom_rule""rule_%s" {
```
ffalor, line 201:
```suggestion
resource "crowdstrike_cloud_posture_custom_rule" "rule_%s" {
```

internal/cloud_posture/custom_rule_resource_test.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func generateMinimalRuleLogicTests(config ruleCustomConfig, ruleName string) []resource.TestStep {
+	var steps []resource.TestStep
+	resourceName := "crowdstrike_cloud_posture_custom_rule.rule" + "_" + ruleName
+
+	for i := range 2 {
+		resourceStep := resource.TestStep{
+			Config: fmt.Sprintf(`
+resource "crowdstrike_cloud_posture_custom_rule""rule_%s" {
```
ffalor, line 304:
```suggestion
resource "crowdstrike_cloud_posture_custom_rule" "rule_%s" {
```

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,984 @@
...
+	}
+
+	plan.AttackTypes = types.SetValueMust(types.StringType, []attr.Value{})
+	for _, attackType := range rule.AttackTypes {
+		plan.AttackTypes, _ = types.SetValue(types.StringType, append(plan.AttackTypes.Elements(), types.StringValue(attackType)))
+		if diags.HasError() {
+			resp.Diagnostics.Append(diags...)
+			return
+		}
+	}
```
ffalor, line 435: we aren't assigning the diags from `SetValue` so our error check is not checking the error.

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+						"auto_remediable": schema.BoolAttribute{
+							Optional:    true,
+							Computed:    true,
+							Description: "Autoremediation enabled for rule",
+						},
+						"domain": schema.StringAttribute{
+							Optional:    true,
+							Computed:    true,
+							Description: "Timestamp of the last Terraform update of the resource.",
+						},
```
ffalor, line 178: Incorrect description

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,984 @@
...
+			},
+			"subdomain": schema.StringAttribute{
+				Required:    true,
+				Description: "Subdomain for the policy rule. Valid values are 'IOM' (Indicators of Misconfiguration) or 'IaC' (Infrastructure as Code). IOM is only supported at this time.",
+				Validators: []validator.String{
+					stringvalidator.OneOf(
+						"IOM",
+						"IAC",
+					),
+				},
```
ffalor, line 258: `IaC` should be `IAC` in the description since a cx may try to use `IaC` as the value due to the description.

It also feels weird to add an option for something that can't be used. Maybe we should either remove this field or only allow `IOM` as a value for now.

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+			Description:     types.StringPointerValue(resource.Description),
+			AutoRemediable:  types.BoolPointerValue(resource.AutoRemediable),
+			Domain:          types.StringPointerValue(resource.Domain),
+			Logic:           types.StringValue(resource.Logic),
+			Name:            types.StringPointerValue(resource.Name),
+			ParentRuleID:    types.StringValue(resource.ParentRuleShortUUID),
+			CloudPlatform:   types.StringValue(resource.Platform),
+			CloudProvider:   types.StringPointerValue(resource.Provider),
+			RemediationInfo: types.StringPointerValue(resource.RuleLogicList[0].RemediationInfo),
+			ResourceType:    types.StringPointerValue(resource.ResourceTypes[0].ResourceType),
```
ffalor, line 394: An example of my other comment around being careful of checking before we try to index. I'm not going to leave a comment on every instance, just wanted to give another example.

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+	Rules         []cloudPostureRulesDataSourceRuleModel `tfsdk:"rules"`
+}
+
+type cloudPostureRulesDataSourceRuleModel struct {
+	UUID      types.String `tfsdk:"uuid"`
+	AlertInfo types.List   `tfsdk:"alert_info"`
+	Controls  []struct {
+		Authority types.String `tfsdk:"authority"`
+		Code      types.String `tfsdk:"code"`
+	} `tfsdk:"controls"`
```
ffalor, line 48: Similiar to the resource, this should be a tf native type instead of a []struct. If there are common models then they can be abstracted out and used in both the data source and resource.

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+					stringvalidator.OneOf(
+						"AWS",
+						"Azure",
+						"GCP",
+					),
+				},
+			},
+			"rule_name": schema.StringAttribute{
+				Optional:    true,
+				Description: "Name of the rule to search for. If no name is defined all rules in cloud provider will be returned.",
```
ffalor, line 123:
```suggestion
				Description: "Name of the rule to search for. If no name is defined all rules in a cloud provider will be returned.",
```

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+			},
+			"rules": schema.ListNestedAttribute{
+				Optional:    true,
```
ffalor, line 126: If this can't be used to filter the api call then it should be only computed. `rules` will be a empty list if nothing is returned from the api. 

Same goes for the other attributes under this, if a cx can't set it in the data source it should be marked as read only (computed)

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+				Computed:    true,
+				Description: "List of cloud posture rules",
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"uuid": schema.StringAttribute{
```
ffalor, line 131: similiar to other comments, lets use `id` i'm not sure why some apis change this to uuid

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+									regexp.MustCompile(`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`),
+									"must be a valid UUID in the format of 7c86a274-c04b-4292-9f03-dafae42bde97",
+								),
+							},
+						},
+						"alert_info": schema.ListAttribute{
+							Optional:    true,
+							Computed:    true,
+							ElementType: types.StringType,
+							Description: "A list of the alert logic and detection criteria for rule violations. Parent value will be used when parent_rule_id is defined.",
```
ffalor, line 145: Description needs to be updated for the data source

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+						},
+						"controls": schema.SetNestedAttribute{
+							Optional:    true,
+							Computed:    true,
+							Description: "Security framework and compliance rule information.",
+							NestedObject: schema.NestedAttributeObject{
+								Attributes: map[string]schema.Attribute{
+									"authority": schema.StringAttribute{
+										Required:    true,
+										Description: "This compliance framework",
```
ffalor, line 155:
```suggestion
										Description: "The compliance framework",
```

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+						},
+						"domain": schema.StringAttribute{
+							Optional:    true,
+							Computed:    true,
+							Description: "Timestamp of the last Terraform update of the resource.",
+						},
+						"logic": schema.StringAttribute{
+							Optional:    true,
+							Computed:    true,
+							Description: "Rego logic for the rule. If this is not defined, then parent_rule_id must be defined.",
```
ffalor, line 182: Double check the other description too, it looks like this description is from the resource and doesn't fit in the data source.

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+	_ context.Context,
+	_ datasource.SchemaRequest,
+	resp *datasource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: utils.MarkdownDescription(
+			"Cloud Posture",
+			"This data source retrieves detailed information about a specific cloud posture rule, including its unique identifier (UUID) and associated attributes.",
+			cloudPostureRuleScopes,
+		),
```
ffalor, line 108: are we required to provide cloud provider? or can this api return all? If it can return all then this data source should support that.

We should also support the other filter options similiar to how we do in the framework data source. IE filter, etc

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+			return nil, diags
+		}
+
+		if notFound, ok := err.(*cloud_policies.QueryRuleInternalServerError); ok {
+			diags.AddError(
+				"Error Querying Rules",
+				fmt.Sprintf("Failed to create rule: %s", *notFound.Payload.Errors[0].Message),
+			)
+			return nil, diags
+		}
```
ffalor, line 324:
```suggestion
		if badRequest, ok := err.(*cloud_policies.QueryRuleBadRequest); ok {
			diags.AddError(
				"Error Querying Rules",
				fmt.Sprintf("Failed to create rule: %s", *badRequest.Payload.Errors[0].Message),
			)
			return nil, diags
		}

		if internalServerError, ok := err.(*cloud_policies.QueryRuleInternalServerError); ok {
			diags.AddError(
				"Error Querying Rules",
				fmt.Sprintf("Failed to create rule: %s", *internalServerError.Payload.Errors[0].Message),
			)
			return nil, diags
		}
```

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+
+		diags.AddError(
+			"Failed to create rule.",
+			fmt.Sprintf("Failed to create rule: %s", err),
+		)
```
ffalor, line 329: this error message needs to be updated

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+		return nil, diags
+	}
+
+	if len(queryPayload.Resources) == 0 {
+		diags.AddWarning(
+			"No Rules Found",
+			"The query returned no rules. Please check your filter criteria.",
+		)
+		return nil, diags
+	}
```
ffalor, line 350: I don't think we need this warning.

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+
+	getRulesResp, err := r.client.CloudPolicies.GetRule(&ruleParams)
+	if err != nil {
+
+		if !strings.Contains(err.Error(), "rule resource doesn't exist") {
+			diags.AddError(
+				"Failed to get rule",
+				fmt.Sprintf("Failed to get rule: %s", err),
+			)
```
ffalor, line 364:
```suggestion
			diags.AddError(
				"Failed to get rules",
				fmt.Sprintf("Failed to get rules: %s", err),
			)
```

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+		}
+		return nil, diags
+	}
+
+	getRulesPayload := getRulesResp.GetPayload()
+
+	if err = falcon.AssertNoError(getRulesPayload.Errors); err != nil {
+		diags.AddError(
+			"Failed to get rules",
+			fmt.Sprintf("Failed to get rule: %s", err.Error()),
```
ffalor, line 374:
```suggestion
			fmt.Sprintf("Failed to get rules: %s", err.Error()),
```

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,426 @@
...
+			})
+		}
+
+		rule.AttackTypes = types.SetValueMust(types.StringType, []attr.Value{})
+		for _, attackType := range resource.AttackTypes {
+			rule.AttackTypes, diags = types.SetValue(types.StringType, append(rule.AttackTypes.Elements(), types.StringValue(attackType)))
+			if diags.HasError() {
+				return nil, diags
+			}
+		}
```
ffalor, line 420: you should be able to do this instead


```suggestion
        attackTypeSet, diags := types.SetValueFrom(ctx, types.StringType, resource.AttackTypes)
        if diags.HasError() {
            return nil, diags
        }
```

internal/cloud_compliance/framework_control_data_source.go
```diff
@@ -0,0 +1,364 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if config.Benchmark.IsNull() && config.FQL.IsNull() {
+		resp.Diagnostics.AddError(
+			"Invalid Configuration",
+			"At least one of 'benchmark' or 'fql' must be defined",
+		)
+	}
```
ffalor, line 181: Is it possible for this endpoint to return everything? if it is then instead of requiring fql or benchmark, we should allow nothing to be provided, and just verify they aren't providing both are the same time.

ffalor (changes requested):
We need to add pagination support also for these get requests. I implement an example here https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/541081e7bb873ab5ce4dd97c79580c9ca289beeb/internal/content_update_policy/content_update_policy_precedence.go#L294

There are a few pagination methods our API's use so implementation varies.

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+var (
+	_ resource.Resource                = &cloudPostureCustomRuleResource{}
+	_ resource.ResourceWithConfigure   = &cloudPostureCustomRuleResource{}
+	_ resource.ResourceWithImportState = &cloudPostureCustomRuleResource{}
+)
```
ffalor:
```suggestion
var (
	_ resource.Resource                		= &cloudPostureCustomRuleResource{}
	_ resource.ResourceWithConfigure   		= &cloudPostureCustomRuleResource{}
	_ resource.ResourceWithImportState 		= &cloudPostureCustomRuleResource{}
	_ resource.ResourceWithConfigValidators = &cloudPostureCustomRuleResource{}
)
```

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+		m.Controls, diags = types.SetValueFrom(
+			ctx,
+			types.ObjectType{AttrTypes: policyControl{}.AttributeTypes()},
+			policyControls,
+		)
+
+		if diags.HasError() {
+			return diags
+		}
+	}
```
ffalor: same issue as bullet point 3 there if state has a value for controls, but the backend returns nothing etc

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+	for _, control := range controls {
+		body.Controls = append(body.Controls, &models.DbmodelsControlReference{
+			Authority: control.Authority.ValueStringPointer(),
+			Code:      control.Code.ValueStringPointer(),
+		})
+	}
+
+	severity := severityToInt64[plan.Severity.ValueString()]
+	body.Severity = severity
```
ffalor:
```suggestion
	body.Severity = severityToInt64[plan.Severity.ValueString()]
```

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+	var diags diag.Diagnostics
+
+	body := &models.CommonUpdateRuleRequest{
+		Description: plan.Description.ValueString(),
+		Name:        plan.Name.ValueString(),
+		UUID:        plan.ID.ValueStringPointer(),
+	}
+
+	severity := severityToInt64[plan.Severity.ValueString()]
+	body.Severity = severity
```
ffalor:
```suggestion
	body.Severity = severityToInt64[plan.Severity.ValueString()]
```

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+		attackTypes := make([]string, 0, len(plan.AttackTypes.Elements()))
+		for _, elem := range plan.AttackTypes.Elements() {
+			if str, ok := elem.(types.String); ok {
+				attackTypes = append(attackTypes, str.ValueString())
+			}
+		}
+		body.AttackTypes = attackTypes
+	} else {
+		body.AttackTypes = []string{}
+	}
```
ffalor: can move `attackTypes := make([]string, 0, len(plan.AttackTypes.Elements()))` outside the if statement and remove the else statement since attackTypes will default to []string{} 

In the if statement use .ElementsAs instead of a for loop, like in my other suggestion

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+
+	m.ID = types.StringPointerValue(rule.UUID)
+	m.Name = types.StringPointerValue(rule.Name)
+	m.Description = types.StringPointerValue(rule.Description)
+	m.Domain = types.StringPointerValue(rule.Domain)
+	m.Subdomain = types.StringPointerValue(rule.Subdomain)
+	m.CloudProvider = types.StringPointerValue(rule.Provider)
+	m.RemediationInfo = convertAlertRemediationInfoToTerraformState(*rule.Remediation)
+
+	m.AlertInfo = convertAlertRemediationInfoToTerraformState(*rule.AlertInfo)
```
ffalor: if these pointers are null this will error. You can update the function to take the pointer and in the function check if null or do the null check before

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+		remediationInfo, diags = convertRemediationInfoToAPIFormat(ctx, plan.RemediationInfo, false)
+		if diags.HasError() {
+			return nil, diags
+		}
+
+		body.RuleLogicList[0].Logic = plan.Logic.ValueString()
+	} else {
+		var parentRule cloudPostureCustomRuleResourceModel
+		var ruleResp *models.ApimodelsRule
+		emptyRemediationInfo := plan.AlertInfo.IsNull() || len(plan.RemediationInfo.Elements()) == 0
```
ffalor: This checks if AlertInfo is null, im assuming it should be checking remediation info

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,540 @@
...
+				ID:             types.StringValue(*resource.UUID),
+				Description:    types.StringPointerValue(resource.Description),
+				AutoRemediable: types.BoolPointerValue(resource.AutoRemediable),
+				Domain:         types.StringPointerValue(resource.Domain),
+				Logic:          types.StringValue(resource.Logic),
+				Name:           types.StringPointerValue(resource.Name),
+				ParentRuleID:   types.StringValue(resource.ParentRuleShortUUID),
+				CloudPlatform:  types.StringValue(resource.Platform),
+				CloudProvider:  types.StringPointerValue(resource.Provider),
+				Severity:       types.StringValue(int64ToSeverity[*resource.Severity]),
```
ffalor: dereferencing a pointer without a nil check

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,540 @@
...
+				return defaultResponse, diags
+			}
+
+			rule.AttackTypes, diags = types.SetValueFrom(ctx, types.StringType, resource.AttackTypes)
+			if diags.HasError() {
+				return defaultResponse, diags
+			}
+
+			if resource.RuleLogicList != nil {
+				rule.RemediationInfo = convertAlertRemediationInfoToTerraformState(resource.RuleLogicList[0].RemediationInfo)
```
ffalor: not checking len here to ensure 0 index is valid

internal/cloud_posture/rules_data_source.go
```diff
@@ -0,0 +1,540 @@
...
+	var diags diag.Diagnostics
+	var filter string
+	limit := int64(500)
+	offset := int64(0)
+	defaultResponse := types.SetValueMust(types.ObjectType{AttrTypes: cloudPostureRulesDataSourceRuleModel{}.AttributeTypes()}, []attr.Value{})
+
+	queryParams := cloud_policies.QueryRuleParams{
+		Context: ctx,
+		Limit:   &limit,
+	}
```
ffalor: offset is never used,

ffalor (changes requested):

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+				MarkdownDescription: "Specific attack types associated with the rule. " +
+					"Note: If `parent_rule_id` is defined, attack types will be inherited from the parent rule and cannot be specified using this field.",
+				ElementType: types.StringType,
+				// Default:     setdefault.StaticValue(types.SetValueMust(types.StringType, []attr.Value{})),
+			},
+			"logic": schema.StringAttribute{
+				Optional: true,
+				MarkdownDescription: "Rego logic for the rule. " +
+					"If this is not defined, then parent_rule_id must be defined. " +
+					"When `parent_rule_id` is defined, `logic` from the parent rule is not visible, but it is used for triggering this rule.",
```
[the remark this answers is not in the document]
gpontejos: Correct. It's currently being excluded from responses. I believe we'll have this as a modifiable field in the future though.

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+		body.AlertInfo, diags = convertAlertInfoToAPIFormat(ctx, plan.AlertInfo, true)
+		if diags.HasError() {
+			return nil, diags
+		}
+
+		body.RemediationInfo, diags = convertRemediationInfoToAPIFormat(ctx, plan.RemediationInfo, true)
+		if diags.HasError() {
+			return nil, diags
+		}
+	}
```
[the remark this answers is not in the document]
gpontejos: Updated to isDuplicateRule

internal/cloud_posture/custom_rule_resource.go
```diff
@@ -0,0 +1,925 @@
...
+
+	if rule.RuleLogicList != nil {
+		m.CloudPlatform = types.StringPointerValue(rule.RuleLogicList[0].Platform)
+	}
+
+	if rule.ResourceTypes != nil {
+		m.ResourceType = types.StringPointerValue(rule.ResourceTypes[0].ResourceType)
+	}
```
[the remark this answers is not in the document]
gpontejos: API folks said these will always be length 1 and the list will remain.

internal/cloud_posture/custom_rule_resource.go
gpontejos: Updated to ensure assignment always occurs.

PR #153, opened 2025-09-22
IT Automation (Falcon for IT)

zantinelli:
Addition of support for the CrowdStrike IT Automation (Falcon for IT) module, enabling management of policies, tasks, task groups, and precedence configuration.

 Resources Added:
  - crowdstrike_it_automation_policy - Manage IT automation policies with execution settings, concurrency limits, and host group assignments
  - crowdstrike_it_automation_default_policy - Configure platform-specific default policies (Windows, Linux, Mac)
  - crowdstrike_it_automation_policy_precedence - Control policy execution order with strict/dynamic enforcement modes
  - crowdstrike_it_automation_task - Create and manage tasks supporting script content, script files, and OSQuery execution
  - crowdstrike_it_automation_task_group - Group tasks for RBAC and organizational purposes

Features:
  - Multi-platform support with platform-specific resource configurations (Windows/Linux vs Mac)
  - Flexible task execution via script content, uploaded script files, or OSQuery
  - Access control with Public/Shared permissions and user assignment
  - Policy precedence management with strict and dynamic enforcement modes
  - Comprehensive validation for configuration consistency and API requirements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ website/node_modules
 *.
 .clinerules/
 terraform-provider-crowdstrike*
+CLAUDE.md
 
 website/vendor
 .clineignore
--- a/internal/fcs/cloud_aws_account.go
+++ b/internal/fcs/cloud_aws_account.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-	"time"
 
 	"github.com/crowdstrike/gofalcon/falcon/client"
 	"github.com/crowdstrike/gofalcon/falcon/client/cloud_aws_registration"
@@ -33,7 +32,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
-	"golang.org/x/exp/maps"
 )
 
 type cloudAWSAccountResource struct {
@@ -541,7 +539,7 @@ func (r *cloudAWSAccountResource) Create(
 		)
 	}
 
-	state.SensorManagement.Enabled = types.BoolValue(*cspmAccount.SensorManagementEnabled)
+	state.SensorManagement.Enabled = types.BoolPointerValue(cspmAccount.SensorManagementEnabled)
 
 	state.DSPM.Enabled = types.BoolValue(cspmAccount.DspmEnabled)
 
@@ -567,11 +565,6 @@ func (r *cloudAWSAccountResource) Create(
 			state.IDP.Enabled = types.BoolValue(true)
 			break
 		}
-	}
-
-	// if IDP or realtime visibility are enabled wait until we get all values back
-	if state.IDP.Enabled.ValueBool() || state.RealtimeVisibility.Enabled.ValueBool() {
-		resp.Diagnostics.Append(r.pollMissingValues(ctx, &state)...)
 	}
 
 	// Set refreshed state
@@ -864,7 +857,7 @@ func (r *cloudAWSAccountResource) Read(
 	} else {
 		state.SensorManagement = &sensorManagementOptions{}
 	}
-	state.SensorManagement.Enabled = types.BoolValue(*cspmAccount.SensorManagementEnabled)
+	state.SensorManagement.Enabled = types.BoolPointerValue(cspmAccount.SensorManagementEnabled)
 
 	if oldState.DSPM != nil {
 		state.DSPM = oldState.DSPM
@@ -1102,7 +1095,7 @@ func (r *cloudAWSAccountResource) Update(
 		)
 	}
 
-	plan.SensorManagement.Enabled = types.BoolValue(*cspmAccount.SensorManagementEnabled)
+	plan.SensorManagement.Enabled = types.BoolPointerValue(cspmAccount.SensorManagementEnabled)
 
 	plan.DSPM.Enabled = types.BoolValue(cspmAccount.DspmEnabled)
 
@@ -1449,91 +1442,5 @@ func (r *cloudAWSAccountResource) ValidateConfig(
 	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
 	if resp.Diagnostics.HasError() {
 		return
-	}
-}
-
-// pollMissingValues waits for missing backend values [eventbus_arn, eventbus_name, cloudtrail_bucket_name] to be set.
-func (r *cloudAWSAccountResource) pollMissingValues(
-	ctx context.Context,
-	config *cloudAWSAccountModel,
-) diag.Diagnostics {
-	var diags diag.Diagnostics
-	missingValues := make(map[string]interface{})
-
-	if config.EventbusArn.ValueString() == "" {
-		missingValues["eventbus_arn"] = nil
-	}
-
-	if config.EventbusName.ValueString() == "" {
-		missingValues["eventbus_name"] = nil
-	}
-
-	if config.CloudTrailBucketName.ValueString() == "" {
-		missingValues["cloudtrail_bucket_name"] = nil
-	}
-
-	if len(missingValues) == 0 {
-		return diags
-	}
-
-	pollCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
-	defer cancel()
-
-	ticker := time.NewTicker(10 * time.Second)
-	defer ticker.Stop()
-
-	for {
-		select {
-		case <-ticker.C:
-			account, diag := r.getCSPMAccount(pollCtx, config.AccountID.ValueString())
-			if diag.HasError() {
-				return diag
-			}
-
-			if _, ok := missingValues["eventbus_arn"]; ok && account.AwsEventbusArn != "" {
-				config.EventbusArn = types.StringValue(account.AwsEventbusArn)
-				delete(missingValues, "eventbus_arn")
-			}
-
-			if _, ok := missingValues["eventbus_name"]; ok && account.EventbusName != "" {
-				config.EventbusName = types.StringValue(account.EventbusName)
-				delete(missingValues, "eventbus_name")
-			}
-
-			if _, ok := missingValues["cloudtrail_bucket_name"]; ok &&
-				account.AwsCloudtrailBucketName != "" {
-				config.CloudTrailBucketName = types.StringValue(account.AwsCloudtrailBucketName)
-				delete(missingValues, "cloudtrail_bucket_name")
-			}
-
-			if len(missingValues) == 0 {
-				return diags
-			}
-
-			fields := strings.Join(maps.Keys(missingValues), ", ")
-			tflog.Debug(
-				pollCtx,
-				fmt.Sprintf("[%s] not yet available, polling again in 10s...", fields),
-			)
-
-		case <-pollCtx.Done():
-			if pollCtx.Err() != context.DeadlineExceeded {
-				return diags
-			}
-
-			if len(missingValues) == 0 {
-				return diags
-			}
-
-			fields := strings.Join(maps.Keys(missingValues), ", ")
-			diags.AddError(
-				"Timed out waiting for missing fields to populate",
-				fmt.Sprintf(
-					"Timed out on create waiting for missing fields [%s] to return from API.",
-					fields,
-				),
-			)
-			return diags
-		}
 	}
 }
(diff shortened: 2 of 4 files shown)


internal/it_automation/it_automation_policy_resource.go
```diff
@@ -0,0 +1,891 @@
...
+		if strings.Contains(strings.ToLower(err.Error()), "not found") ||
+			strings.Contains(err.Error(), "404") {
+			tflog.Warn(
+				ctx,
+				fmt.Sprintf(notFoundRemoving, fmt.Sprintf("%s %s", itAutomationPolicy, policyID)),
+				map[string]any{"error": err.Error()},
+			)
+			resp.State.RemoveResource(ctx)
+			return
+		}
```
ffalor: we don't have to remove the resource from state on delete. Terraform core handles this. 
```suggestion
		if strings.Contains(strings.ToLower(err.Error()), "not found") ||
			strings.Contains(err.Error(), "404") {
			return
		}
```

internal/it_automation/it_automation_policy_resource.go
```diff
@@ -0,0 +1,891 @@
...
+		_, verifyDiags := getItAutomationPolicy(ctx, r.client, policyID)
+		if verifyDiags.HasError() {
+			for _, diag := range verifyDiags {
+				if diag.Summary() == policyNotFoundErrorSummary {
+					resp.State.RemoveResource(ctx)
+					return
+				}
+			}
+		}
+
```
ffalor:
```suggestion
		if verifyDiags.HasError() {
			for _, diag := range verifyDiags {
				if diag.Summary() == policyNotFoundErrorSummary {
					return
				}
			}
		}

```

internal/it_automation/it_automation_policy_resource.go
```diff
@@ -0,0 +1,891 @@
...
+	ok, err := r.client.ItAutomation.ITAutomationCreatePolicy(params)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating IT automation policy",
+			"Could not create policy, unexpected error: "+err.Error(),
+		)
+		return
+	}
+
+	policy := ok.Payload.Resources[0]
```
ffalor: Add proper nil checks before accessing nested fields

internal/it_automation/it_automation_policy_resource.go
```diff
@@ -0,0 +1,891 @@
...
+				if err != nil {
+					resp.Diagnostics.AddError(
+						"Error updating IT automation policy host groups",
+						"Could not update host groups for policy ID: "+policyID+", error: "+err.Error(),
+					)
+					return
+				}
+			}
+		}
+	}
```
ffalor: does this endpoint also remove anything not in the list? We have action assign, but happens when the backend resource has host groups that need to be removed?

internal/it_automation/it_automation_policy_resource.go
```diff
@@ -0,0 +1,891 @@
...
+					t.MemoryAllocation = types.Int32Value(r.MemoryAllocation)
+				}
+				if r.MemoryAllocationUnit != "" {
+					t.MemoryAllocationUnit = types.StringValue(r.MemoryAllocationUnit)
+				}
+			}
+		}
+	}
+
+	t.LastUpdated = types.StringValue(time.Now().Format(timeFormat))
```
ffalor: use	`utils.GenerateUpdateTimestamp()`

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+	MemoryAllocation                types.Int32  `tfsdk:"memory_allocation"`
+	MemoryAllocationUnit            types.String `tfsdk:"memory_allocation_unit"`
+	MemoryPressureLevel             types.String `tfsdk:"memory_pressure_level"`
+	Platform                        types.String `tfsdk:"platform"`
+}
+
+func (t *itAutomationDefaultPolicyResourceModel) wrap(
+	policy models.ItautomationPolicy,
+) {
+	currentDescription := t.Description
```
ffalor: same feedback, .wrap should update the model with the new remote state so if description returns nil then it should be updated to nil and not the current description. Check the other properties.

ffalor (changes requested):
submitting what i have so far

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+					t.MemoryAllocation = types.Int32Value(r.MemoryAllocation)
+				}
+				if r.MemoryAllocationUnit != "" {
+					t.MemoryAllocationUnit = types.StringValue(r.MemoryAllocationUnit)
+				}
+			}
+		}
+	}
+
+	t.LastUpdated = types.StringValue(time.Now().Format(timeFormat))
```
ffalor: .wrap is called in read, since there are no updates on read, the create and update methods should set this.

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+	}
+
+	updatedPolicy, diags := getItAutomationPolicy(ctx, r.client, policyID)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	plan.wrap(*updatedPolicy)
+	plan.LastUpdated = types.StringValue(time.Now().Format(timeFormat))
```
ffalor: lets use the utils function for creating the timestamp

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+				Description: "Name of the default policy. This is read-only as default policy names cannot be changed.",
+			},
+			"description": schema.StringAttribute{
+				Optional:    true,
+				Computed:    true,
+				Description: "Description of the default policy.",
+				Validators: []validator.String{
+					stringvalidator.LengthAtLeast(1),
+				},
+			},
```
ffalor: By marking description as computed we are diverging from how most other resources handle description. Other descriptions are generally marked as required because their APIs do not support setting them to an empty value. APIs that allow the description to be set to null/empty are marked as Optional. 

This being computed would mean if no description is provided in the config, we will need to use the API value.

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+		}
+
+		// get updated policy after configuration changes.
+		updatedPolicy, diags := getItAutomationPolicy(ctx, r.client, *policy.ID)
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
+		plan.wrap(*updatedPolicy)
```
ffalor:
```
./internal/it_automation/it_automation_default_policy_resource.go:374:14: error: Potential nil panic detected. Observed nil flow from source to dereference point:
        - it_automation/shared.go:244:9: literal `nil` returned from `getItAutomationPolicy()` in position 0
        - it_automation/it_automation_default_policy_resource.go:374:14: result 0 of `getItAutomationPolicy()` dereferenced via the assignment(s):
                - `getItAutomationPolicy(...)` to `updatedPolicy` at it_automation/it_automation_default_policy_resource.go:368:3
```

internal/it_automation/it_automation_policy_resource.go
ffalor:
```
./internal/it_automation/it_automation_policy_resource.go:422:12: error: Potential nil panic detected. Observed nil flow from source to dereference point:
        - it_automation/it_automation_client.go:250:10: literal `nil` returned from `ITAutomationCreatePolicy()` in position 0
        - it_automation/it_automation_client.go:63:2: returned as result 0 from interface method `ClientService.ITAutomationCreatePolicy()` (implemented by `Client.ITAutomationCreatePolicy()`)
        - it_automation/it_automation_policy_resource.go:422:12: result 0 of `ITAutomationCreatePolicy()` accessed field `Payload` via the assignment(s):
                - `r.client.ItAutomation.ITAutomationCreatePolicy(...)` to `ok` at it_automation/it_automation_policy_resource.go:413:2
```

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+				tflog.Warn(
+					ctx,
+					fmt.Sprintf(notFoundRemoving, fmt.Sprintf("IT Automation Default Policy %s", policyID)),
+				)
+				resp.State.RemoveResource(ctx)
+				return
+			}
+		}
+		return
+	}
```
ffalor: We append diags into resp before we check if there are any errors that should be ignored. This will result in terraform throwing an error even if the error is just a `policyNotFoundErrorSummary`

Something like this should allow us to return early if the error is a `policyNotFoundErrorSummary` else we append to `resp` and return the errors for terrform core to report.
```suggestion
	policy, diags := getItAutomationPolicy(ctx, r.client, policyID)

	if diags.HasError() {
		// manually parse diagnostic errors.
		// helper functions return standardized diagnostics for consistency.
		// this is due to some IT Automation endpoints not returning structured/generic 404s.
		for _, d := range resp.Diagnostics.Errors() {
			if d.Summary() == policyNotFoundErrorSummary {
				tflog.Warn(
					ctx,
					fmt.Sprintf(notFoundRemoving, fmt.Sprintf("IT Automation Default Policy %s", policyID)),
				)
				resp.State.RemoveResource(ctx)
				return
			}
		}

        resp.Diagnostics.Append(diags...)
		return
	}
```

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+				Validators: []validator.String{
+					stringvalidator.OneOf("MB", "GB"),
+				},
+			},
+			"memory_pressure_level": schema.StringAttribute{
+				Optional:    true,
+				Description: "Sets memory pressure level to control system resource allocation during task execution.",
+				Validators: []validator.String{
+					stringvalidator.OneOf("Low", "Medium", "High"),
+				},
```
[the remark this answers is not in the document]
ffalor: I didn't realize at the time that some of these are platform specific so they are required to be optional and then use validateconfig to require them like you did. Question, when I go to the console the options are all greyed out, can they only be changed by the API? 

The above issue with execution timeout is still valid though.

internal/it_automation/it_automation_default_policy_resource.go
```diff
@@ -0,0 +1,689 @@
...
+				Description: "Identifier for the default policy.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"last_updated": schema.StringAttribute{
+				Computed:    true,
+				Description: "Timestamp of the last Terraform update of the resource.",
+			},
+			"platform": schema.StringAttribute{
```
ffalor: can we use `platform_name` since we have other resources already using `platform_name`

internal/it_automation/it_automation_policy_resource.go
ffalor: You should checkout what I did in sensor update policy to handle this


```go
	hostGroupsToAdd, hostGroupsToRemove, diags := utils.SetIDsToModify(
		ctx,
		plan.HostGroups,
		state.HostGroups,
	)
	resp.Diagnostics.Append(diags...)
		return
	}

	if len(hostGroupsToAdd) != 0 {
		err := r.updateHostGroups(
			ctx,
			hostgroups.AddHostGroup,
			hostGroupsToAdd,
			plan.ID.ValueString(),
		)
		if err != nil {
			resp.Diagnostics.AddError(
				"Error updating CrowdStrike sensor update policy",
				fmt.Sprintf(
					"Could not add host groups: (%s) to policy with id: %s \n\n %s",
					strings.Join(hostGroupsToAdd, ", "),
					plan.ID.ValueString(),
					err.Error(),
				),
			)
			return
		}
	}

	if len(hostGroupsToRemove) != 0 {
		err := r.updateHostGroups(
			ctx,
			hostgroups.RemoveHostGroup,
			hostGroupsToRemove,
			plan.ID.ValueString(),
		)
		if err != nil {
			resp.Diagnostics.AddError(
				"Error updating CrowdStrike sensor update policy",
				fmt.Sprintf(
					"Could not remove host groups: (%s) from policy with id: %s \n\n %s",
					strings.Join(hostGroupsToAdd, ", "),
					plan.ID.ValueString(),
					err.Error(),
				),
			)
			return
		}
	}
```

ffalor (changes requested):
some comments may apply to multiple resources, but for time sake I didn't duplicate them. For example the comment around using platform_name instead of platform

PR #169, opened 2025-10-30
Feature/adar/cspg 67584/vul scanning tf provider

AdarShakedCrowdstrike:
This PR adds comprehensive support CrowdStrike Vulnerability Scanning configuration in
  the AWS cloud account resource, introduces a new computed field for agentless scanning role
  resolution.

Main changes:
* New vulnerability_scanning block with enabled and optional role_name parameters.
* Computed fields for vulnerability_scanning_role_arn and vulnerability_scanning_role_name ,agentless_scanning_role_name.
 * Data source support with all vulnerability scanning fields available in crowdstrike_cloud_aws_account data source

### DIFF
--- a/internal/it_automation/shared_test.go
+++ b/internal/it_automation/shared_test.go
@@ -290,7 +290,7 @@ func deleteUser(
 	_, err := falconClient.UserManagement.DeleteUserV1(params)
 
 	if err != nil {
-		if _, ok := err.(*user_management.DeleteUserNotFound); ok {
+		if strings.Contains(err.Error(), "status 404") {
 			return nil
 		}
 		return fmt.Errorf("DeleteUserV1 failed: %w", err)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/crowdstrike/terraform-provider-crowdstrike
 go 1.24.0
 
 require (
-	github.com/crowdstrike/gofalcon v0.18.0
+	github.com/crowdstrike/gofalcon v0.17.1-0.20251022212655-073422a53e82
```
AdarShakedCrowdstrike: update package value after cspm api is live in prod

internal/fcs/utils.go
```diff
@@ -9,3 +14,22 @@ func getRoleNameFromArn(arn string) string {
 	}
 	return ""
 }
+
+// resolveAgentlessScanningRoleName retrieves agentless scanning role name from CSPM account.
+// DSPM role takes precedence over vulnerability scanning role.
+func resolveAgentlessScanningRoleName(cspmAccount *models.DomainAWSAccountV2) (string, error) {
```
AdarShakedCrowdstrike: @ffalor atm dspm role name always return a value even if disabled. I choose to return values only if enabled, 
if customer not populating the enabled to the provide his tf will fail (for example in the single account, the user sets `enable_dspm` var to true, sends it to the module but not to the provider. it will fail

internal/fcs/account_settings.go
```diff
@@ -0,0 +1,29 @@
+package fcs
+
+import (
+	"fmt"
+
+	"github.com/crowdstrike/gofalcon/falcon/models"
+	"github.com/mitchellh/mapstructure"
```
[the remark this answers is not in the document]
ffalor: `Settings interface{} `json:"settings,omitempty"``

Settings is a interface because of the API spec, does settings need to be an interface? If it has a defined schema then the api spec should be updated.

internal/fcs/cloud_aws_account.go
```diff
@@ -1443,4 +1583,17 @@ func (r *cloudAWSAccountResource) ValidateConfig(
 	if resp.Diagnostics.HasError() {
 		return
 	}
+
+	// Validate DSPM and vulnerability scanning role name consistency
+	if config.DSPM != nil && config.DSPM.Enabled.ValueBool() && config.VulnerabilityScanning != nil && config.VulnerabilityScanning.Enabled.ValueBool() {
```
ffalor: `if config.DSPM != nil`

In validateconfig we need to check if a value is known before we do any validation on it.

If you want to check if a value is known and not null then you can use the helper `utils.IsKnown()`

`if utils.IsKnown(config.DSPM) && config.DSPM.Enabled.ValueBool() && utils.IsKnown(config.VulnerabilityScanning) && config.VulnerabilityScanning.Enabled.ValueBool()`

Double check you are making sure a value isn't unknown with `.IsUnknown()` ie: `config.DSPM.RoleName.IsUnknown()`

internal/fcs/utils.go
```diff
@@ -9,3 +14,22 @@ func getRoleNameFromArn(arn string) string {
...
+	if cspmAccount.DspmEnabled {
+		return getRoleNameFromArn(cspmAccount.DspmRoleArn), nil
+	}
+
+	// Try fallback to vulnerability scanning role if DSPM is not enabled
```
ffalor: The commented behavior about how this role is chosen should be documented in the attribute description in the schema.

internal/fcs/account_settings.go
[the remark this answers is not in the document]
ffalor: Looks like the replacement of that repo is https://github.com/go-viper/mapstructure

internal/fcs/modifiers.go
```diff
@@ -7,27 +7,28 @@ import (
...
 }
 
-func (m dspmRoleArnModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
+func (m roleArnStateModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
 	// Do nothing if there is no state value.
 	if req.StateValue.IsNull() {
 		return
 	}
 
-	// Do nothing if there is a known planned value.
```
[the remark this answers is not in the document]
AdarShakedCrowdstrike: It creates a bug when the plan value is known but the response form the API is different. it actually happened to me few times while adding and removing dspm/vul section.

internal/fcs/account_settings.go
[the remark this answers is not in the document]
AdarShakedCrowdstrike: Yea I know its an interface because of the API spec, Max told me they would like to keep it generic. I would love to have type for this  field from the API spec. I will take it with him and @ashoksankarh-cs.Rr the package, I will update.

internal/fcs/cloud_aws_account.go
ffalor: thank you, i didn't check the type because all models should only contain terraform types so sadly this being a struct is a bug in itself that will cause runtime errors if someone ever uses variables in a way that results in the plan trying to set dspm to unknown.

A example of this I fixed recently in prevention policy: https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/46bacd5a9ad183f0e37765466bfa09cb4f92b18c/internal/prevention_policy/windows_test.go#L73-L114

AdarShakedCrowdstrike: re Double check you are making sure a value isn't unknown with .IsUnknown() ie: config.DSPM.RoleName.IsUnknown().
Done and added tests

internal/fcs/utils.go
AdarShakedCrowdstrike: Done

internal/fcs/modifiers.go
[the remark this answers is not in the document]
AdarShakedCrowdstrike: yea this is the error:
  Once you share the error details, I can help identify the root cause and provide a solution!

> │ Error: Provider produced inconsistent result after apply
  │
  │ When applying changes to crowdstrike_cloud_aws_account.this, provider
  "provider[\"registry.terraform.io/crowdstrike/crowdstrike\"]" produced an unexpected new
  value: .vulnerability_scanning_role_arn: was cty.StringVal(""), but now
  │ cty.StringVal("arn:aws:iam::654654198549:role/cs-agentless-scanning-integration-cspm").
  │
  │ This is a bug in the provider, which should be reported in the provider's own issue
  tracker.
  ╵
  ╷
  │ Error: Provider produced inconsistent result after apply
  │
  │ When applying changes to crowdstrike_cloud_aws_account.this, provider
  "provider[\"registry.terraform.io/crowdstrike/crowdstrike\"]" produced an unexpected new
  value: .vulnerability_scanning_role_name: was cty.StringVal(""), but now
  │ cty.StringVal("cs-agentless-scanning-integration-cspm").
  │
  │ This is a bug in the provider, which should be reported in the provider's own issue
  tracker.

JessicaKane213 (approved):

go.mod
redhatrises: @AdarShakedCrowdstrike please drop the gofalcon version change from go.mod. This should use the latest gofalcon version that is in the go.mod of this project which is `v0.18.1-0.20251101020249-526ce85b6da7`

AdarShakedCrowdstrike: ofc, @ffalor generated a branch with changes which are not in prod yet so I can develop. When  the swagger will be in production I will bump it to the latest version of gofalcon.

Avatichi (approved):

internal/fcs/cloud_aws_account_test.go
```diff
@@ -156,33 +281,222 @@ func TestAccCloudAwsAccountResourceMinimal(t *testing.T) {
...
+					resource.TestCheckResourceAttr(fullResourceName, "vulnerability_scanning_role_name", testVulnRoleName),
+					resource.TestCheckResourceAttr(fullResourceName, "agentless_scanning_role_name", testVulnRoleName),
+				),
+			},
+		},
+	})
+}
+
+func TestAccCloudAwsAccountResourceVulnerabilityScanningNoRoleName(t *testing.T) {
+	t.Skip("skip until CSPM fix for vulnerability scanning role response")
```
ffalor: What is this? Why are we adding a test and skipping it? What's the issue that needs to be fixed?

AdarShakedCrowdstrike: this was fixed by Max, I will rely on new field and not on the setting so I can enable it.

AdarShakedCrowdstrike: Enabled the test

PR #176, opened 2025-11-12
feat: add s3 log ingestion method params

mphillips02-cs:
These changes implement AWS API support for S3-based log ingestion as an alternative to the default EventBridge method for AWS CloudTrail logs.


internal/fcs/cloud_aws_account.go
```diff
@@ -293,18 +298,61 @@ func (r *cloudAWSAccountResource) Schema(
 						Default:     booldefault.StaticBool(true),
 						Description: "Set to true if a CloudTrail already exists",
 					},
+					"log_ingestion_method": schema.StringAttribute{
```
ffalor: Question: these attributes are in the `realtime_visibility` block, does the `log_ingestion` prefix add clarity? Would we ever have attributes here not related to log ingestion or attributes with similar names that would require us to call out ones that are related to log ingestion?

internal/fcs/cloud_aws_account.go
```diff
@@ -682,6 +730,24 @@ func (r *cloudAWSAccountResource) createCloudAccount(
 		ResourceNamePrefix: model.ResourceNamePrefix.ValueString(),
 		ResourceNameSuffix: model.ResourceNameSuffix.ValueString(),
 	}
+	// Add S3 log ingestion fields if realtime visibility is configured
+	if model.RealtimeVisibility != nil {
+		if !model.RealtimeVisibility.LogIngestionMethod.IsNull() {
```
ffalor: if we are forcing a default value of `""` `LogIngestionMethod` etc will never be null.

internal/fcs/cloud_aws_account.go
```diff
@@ -1443,4 +1647,74 @@ func (r *cloudAWSAccountResource) ValidateConfig(
 	if resp.Diagnostics.HasError() {
 		return
 	}
+
+	// Validate S3 log ingestion requirements
+	if config.RealtimeVisibility != nil && !config.RealtimeVisibility.LogIngestionMethod.IsNull() {
```
ffalor: the main thing here is to check if the values are known before validation.

If `config.RealtimeVisibility.LogIngestionMethod` is unknown then we shouldn't do a validation check. The same behavior applies to the other attributes.

ValidateConfig is ran multiple times, sooner or later the value will be known.

internal/fcs/cloud_aws_account.go
ffalor: same feedback applies to the implementation added in the other methods

internal/fcs/cloud_aws_account.go
```diff
@@ -1443,4 +1647,74 @@ func (r *cloudAWSAccountResource) ValidateConfig(
...
+					"Missing required field",
+					"log_ingestion_sns_topic_arn is required when log_ingestion_method is 's3'",
+				)
+			}
+
+			// Validate SNS topic ARN format
+			if !config.RealtimeVisibility.LogIngestionSnsTopicArn.IsNull() {
+				snsArn := config.RealtimeVisibility.LogIngestionSnsTopicArn.ValueString()
+				if snsArn != "" {
+					snsArnPattern := regexp.MustCompile(`^arn:(aws|aws-us-gov|aws-cn):sns:[a-z0-9-]+:[0-9]+:[a-zA-Z0-9_-]+$`)
```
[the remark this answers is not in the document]
ffalor: The same for for bucket name and kms format

internal/fcs/cloud_aws_account.go
mphillips02-cs: i think having log_ingestion as a prefix is valuable a) term real-time visibility does not by itself to me immediately scream log ingestion, so the prefix clarifies what the method field actually is about so that it's not super opaque and b) in the future some other mechanism will be grouped with real-time visibility and detection and in that case this is future-compatible by being specific

internal/fcs/cloud_aws_account.go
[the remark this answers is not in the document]
mphillips02-cs: gotcha, shall do

internal/fcs/cloud_aws_account.go
```diff
@@ -293,18 +298,61 @@ func (r *cloudAWSAccountResource) Schema(
...
+						Default:     stringdefault.StaticString(""),
+						Description: "S3 bucket name for CloudTrail log ingestion when log_ingestion_method is 's3'. Required when using S3 method",
+					},
+					"log_ingestion_sns_topic_arn": schema.StringAttribute{
+						Optional:    true,
+						Computed:    true,
+						Default:     stringdefault.StaticString(""),
+						Description: "SNS topic ARN for S3 CloudTrail log notifications when log_ingestion_method is 's3'. Required when using S3 method",
+					},
+					"log_ingestion_s3_bucket_prefix": schema.StringAttribute{
```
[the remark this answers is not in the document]
mphillips02-cs: yes, there is a sensible limit we can enforce. will move the other validation here and adding that length check

internal/fcs/cloud_aws_account.go
```diff
@@ -293,18 +298,61 @@ func (r *cloudAWSAccountResource) Schema(
...
+						Optional:    true,
+						Computed:    true,
+						Default:     stringdefault.StaticString(""),
+						Description: "Optional S3 bucket prefix for CloudTrail logs when log_ingestion_method is 's3'",
+					},
+					"log_ingestion_kms_key_arn": schema.StringAttribute{
+						Optional:    true,
+						Computed:    true,
+						Default:     stringdefault.StaticString(""),
+						Description: "Optional KMS key ARN for S3 bucket encryption when log_ingestion_method is 's3'",
```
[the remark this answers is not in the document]
mphillips02-cs: based on our conversation, i think these should not be computed, and thus will removed the computed: true field

internal/fcs/cloud_aws_account.go
mphillips02-cs: hmm, not sure how to use `RequiredTogether` here. i understand how and why we'd include `log_ingestion_s3_bucket_name` and `log_ingestion_sns_topic_arn` in `RequiredTogether`, but how would i write that if i only wanted to have those two `RequiredTogether` if `log_ingestion_method` = `s3` ?

internal/fcs/cloud_aws_account.go
```diff
@@ -940,6 +1023,56 @@ func (r *cloudAWSAccountResource) Read(
 		)
 	}
 
+	// Update S3 log ingestion fields from API response settings
+	// All APIs now use period notation consistently
+	state.RealtimeVisibility.LogIngestionMethod = types.StringValue("eventbridge")
+	state.RealtimeVisibility.LogIngestionS3BucketName = types.StringValue("")
+	state.RealtimeVisibility.LogIngestionSnsTopicArn = types.StringValue("")
+	state.RealtimeVisibility.LogIngestionS3BucketPrefix = types.StringValue("")
+	state.RealtimeVisibility.LogIngestionKmsKeyArn = types.StringValue("")
```
ffalor: I think the new default is StringNull

(similiar feedback in the other places this is done)

internal/fcs/cloud_aws_account.go
```diff
@@ -1552,6 +1753,31 @@ func (r *cloudAWSAccountResource) ValidateConfig(
 		return
 	}
 
+	// Validate S3 log ingestion requirements
+	if config.RealtimeVisibility != nil && !config.RealtimeVisibility.LogIngestionMethod.IsNull() {
```
ffalor: `.IsNull` will return false if the value is unknown which means the if statement could pass when `LogIngestionMethod` is unknown.
```suggestion
	if config.RealtimeVisibility != nil && utils.IsKnown(config.RealtimeVisibility.LogIngestionMethod) {
```

`utils.IsKnown` will return true if a value IS NOT unknown and IS NOT null.

internal/fcs/cloud_aws_account.go
```diff
@@ -1552,6 +1753,31 @@ func (r *cloudAWSAccountResource) ValidateConfig(
...
+		method := config.RealtimeVisibility.LogIngestionMethod.ValueString()
+
+		if method == "s3" {
+			if config.RealtimeVisibility.LogIngestionS3BucketName.IsNull() ||
+				!utils.IsKnown(config.RealtimeVisibility.LogIngestionS3BucketName) {
```
ffalor: `.IsNull` is enough here, I was under the impression that `.IsNull()` could return true when a value is unknown, but that was incorrect.

Looking at `.IsNull` implementation, I realized that a attribute can never be both `.IsNull` and `.IsUnknown` because they are completely different states
```go
return s.state == attr.ValueStateNull
```
```go
const (
	// ValueStateNull represents a value which is null.
	//
	// This value is 0 so it is the zero-value for types implementations.
	ValueStateNull ValueState = 0

	// ValueStateUnknown represents a value which is unknown.
	ValueStateUnknown ValueState = 1

	// ValueStateKnown represents a value which is known (not null or unknown).
	ValueStateKnown ValueState = 2
)
```

```suggestion
			if config.RealtimeVisibility.LogIngestionS3BucketName.IsNull() {
```


tldr `.IsNull` will only return true when a value is known and null which is what we want here

ffalor (changes requested):

internal/fcs/cloud_aws_account.go
```diff
@@ -303,18 +309,86 @@ func (r *cloudAWSAccountResource) Schema(
...
+								stringvalidator.All(
+									stringvalidator.RegexMatches(
+										regexp.MustCompile(`^[a-z0-9][a-z0-9.-]*[a-z0-9]$`),
+										"must be 3-63 characters, contain only lowercase letters, numbers, dots, and hyphens, and start/end with alphanumeric characters",
+									),
+									stringvalidator.LengthAtLeast(3),
+									stringvalidator.LengthAtMost(63),
+								)),
+						},
+					},
```
ffalor: LMK if I am missing something, but this value should either be null or atleast 3-63 characters so there is no need too have a OR check for 0 characters.
```suggestion
					"log_ingestion_s3_bucket_name": schema.StringAttribute{
						Optional:    true,
						Description: "S3 bucket name for CloudTrail log ingestion when log_ingestion_method is 's3'. Required when using S3 method",
						Validators: []validator.String{
							stringvalidator.All(
								stringvalidator.RegexMatches(
									regexp.MustCompile(`^[a-z0-9][a-z0-9.-]*[a-z0-9]$`),
									"must be 3-63 characters, contain only lowercase letters, numbers, dots, and hyphens, and start/end with alphanumeric characters",
								),
								stringvalidator.LengthAtLeast(3),
								stringvalidator.LengthAtMost(63),
							),
						},
					},
```

ffalor (changes requested):

ffalor (changes requested):
It looks like the acc tests are failing because the api is not returning s3.log.ingestion.kms.key.arn and s3.log.ingestion.bucket.prefix

Im testing in us-2 so I am not sure if that is part of it.

ffalor (approved):

PR #181, opened 2025-11-15
AWS registration: Add support to set rtvd regions

ashoksankarh-cs:
Add support to set rtvd regions. This will help update the regions in Falcon Backend

### DIFF
--- a/internal/fcs/cloud_aws_account_test.go
+++ b/internal/fcs/cloud_aws_account_test.go
@@ -91,11 +91,9 @@ resource "crowdstrike_cloud_aws_account" "test" {
   }
   dspm = {
     enabled       = true
-    regions = ["us-east-1", "eu-west-1"]
   }
   vulnerability_scanning = {
     enabled                          = true
-    regions = ["us-east-1", "ap-southeast-1"]
   }
 }
 `, account)
@@ -112,11 +110,9 @@ resource "crowdstrike_cloud_aws_account" "test" {
   }
   dspm = {
     enabled       = true
-    regions = ["us-west-2", "eu-west-1"]
   }
   vulnerability_scanning = {
-    enabled                          = true
-    regions = ["us-east-1"]
+    enabled       = true
   }
 }
 `, account)
@@ -228,7 +224,6 @@ resource "crowdstrike_cloud_aws_account" "test" {
   account_id = "%s"
   dspm = {
     enabled = true
-    regions = ["us-east-1"]
   }
 }
 `, account)


internal/fcs/cloud_aws_account.go
```diff
@@ -303,18 +306,32 @@ func (r *cloudAWSAccountResource) Schema(
 						Default:     booldefault.StaticBool(true),
 						Description: "Set to true if a CloudTrail already exists",
 					},
+					"rtvd_regions": schema.ListAttribute{
+						Optional:    true,
+						Computed:    true,
+						Description: "List of AWS regions for Real-Time Visibility and Detection",
```
ffalor: The parent object gives the context so let's rename this to regions 

The same for the others.

internal/fcs/cloud_aws_account.go
```diff
@@ -764,6 +809,32 @@ func (r *cloudAWSAccountResource) createCloudAccount(
 		ResourceNamePrefix: model.ResourceNamePrefix.ValueString(),
 		ResourceNameSuffix: model.ResourceNameSuffix.ValueString(),
 	}
+
+	// Add regions if specified
+	if model.RealtimeVisibility != nil && !model.RealtimeVisibility.RtvdRegions.IsNull() && !model.RealtimeVisibility.RtvdRegions.IsUnknown() {
```
ffalor: For brevity utils.IsKnown can be used to verify a value is not null and is not unknown.

internal/fcs/cloud_aws_account.go
```diff
@@ -764,6 +809,32 @@ func (r *cloudAWSAccountResource) createCloudAccount(
...
+	}
+
+	if model.VulnerabilityScanning != nil && !model.VulnerabilityScanning.VulnerabilityScanningRegions.IsNull() && !model.VulnerabilityScanning.VulnerabilityScanningRegions.IsUnknown() {
+		var vulnRegions []string
+		diags.Append(model.VulnerabilityScanning.VulnerabilityScanningRegions.ElementsAs(ctx, &vulnRegions, false)...)
+		if !diags.HasError() {
+			createAccount.VulnerabilityScanningRegions = vulnRegions
+		}
+	}
+
```
ffalor: If ElementsAs fails we should return an error instead of silently continuing or we will run into state issues later. 

To reduce the chance of failures to basically zero we should also add validators in the schema for these new attributes. 

For example a minimum count on the list and a minimum length on the strings in the list. 

This will prevent the list from ever being an empty list, or a list with empty strings etc.

ashoksankarh-cs: We can probably add a minimum count to avoid entering an empty list. But the BE processes all those gracefully. Null, empty list or list of region strings. Which is why I didnt bother adding more validations here. Let me know your comments

ffalor: The minimum count is more for terraform and less for the backend. If we do not have a minimum count then the following are valid:

- regions = []
- regions = ["us-east-1", nil, "us-west-1"]
- omitting regions all together

The issue I was referring to is when we have to write values to state. We will need to conditionally decide how to write to state. If the api returns an empty list, then we have to check the config and either write null or an empty list depending on the config. Requiring a minimum count means regions = [] is never valid so when we write to state we either write a list of regions or we write a null list. It just makes that part of the provider easier by removing that option. 

The validation on minimum length for the string is to prevent lists being passed with a null value.

- regions = ["us-east-1", nil, "us-west-1"]

internal/fcs/cloud_aws_account.go
```diff
@@ -303,18 +306,32 @@ func (r *cloudAWSAccountResource) Schema(
 						Default:     booldefault.StaticBool(true),
 						Description: "Set to true if a CloudTrail already exists",
 					},
+					"rtvd_regions": schema.ListAttribute{
+						Optional:    true,
+						Computed:    true,
```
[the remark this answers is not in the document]
ashoksankarh-cs: It still gets updated to the settings

ffalor (approved):

ffalor (dismissed):

internal/fcs/cloud_aws_account.go
```diff
@@ -1739,3 +1811,73 @@ func (r *cloudAWSAccountResource) ValidateConfig(
 		}
 	}
 }
+
+// parseRegionsFromSettings parses region information from the cloudAccount Settings map
+// and populates the appropriate region fields in the state model.
+func parseRegionsFromSettings(ctx context.Context, settings interface{}, state *cloudAWSAccountModel) diag.Diagnostics {
```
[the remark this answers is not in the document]
ffalor: @AdarShakedCrowdstrike In (https://github.com/CrowdStrike/terraform-provider-crowdstrike/pull/181/commits/4a43be1ca9912cd10d7b675c7ae5831a2a13ccd4) I moved this to its own file and added log ingestion settings, the code is written with the assumption that the other registrations will need similar settings lmk if that is wrong.

The change also fixes a bug that could cause state inconsistency and cleans up some already existing logic.

internal/fcs/cloud_aws_account.go
```diff
@@ -1124,21 +1125,6 @@ func (r *cloudAWSAccountResource) Read(
 		}
 		state.ResourceNamePrefix = types.StringValue(cloudAccount.ResourceNamePrefix)
 		state.ResourceNameSuffix = types.StringValue(cloudAccount.ResourceNameSuffix)
-
-		// Parse regions from Settings map
-		diags := parseRegionsFromSettings(ctx, cloudAccount.Settings, &state)
```
AdarShakedCrowdstrike: dont we need to parse settings and save then in state when reading account data?

cs-wayne-hsieh (dismissed):

ffalor (approved):

PR #191, opened 2025-11-19
feat: Add cloud_risks data source for Falcon Cloud Security Cloud risks findings

morinies:
Adds a new data source for querying cloud risks (findings) from Falcon Cloud Security.

This PR introduces:
- New `crowdstrike_cloud_risks` data source
- Full support for FQL filtering and sorting
- Automatic pagination to fetch all results
- Comprehensive attribute coverage for cloud risk details
- Acceptance tests for validation

### DIFF
--- a/internal/cloud_security/cloud_risk_findings_data_source.go
+++ b/internal/cloud_security/cloud_risk_findings_data_source.go
@@ -7,12 +7,15 @@ import (
 	"github.com/crowdstrike/gofalcon/falcon"
 	"github.com/crowdstrike/gofalcon/falcon/client"
 	"github.com/crowdstrike/gofalcon/falcon/client/cloud_security"
+	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/framework/validators"
 	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/scopes"
+	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/tferrors"
 	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/utils"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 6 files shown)


internal/cloud_security/cloud_risk_findings_data_source.go
```diff
@@ -0,0 +1,429 @@
...
+		}
+
+		// Check pagination - stop if we've fetched all results
+		if payload.Meta == nil || payload.Meta.Pagination == nil || payload.Meta.Pagination.Total == nil {
+			diags.AddError(
+				"Error Fetching Cloud Risks",
+				"The API returned a response without pagination metadata. Cannot safely paginate results.",
+			)
+			return defaultResponse, diags
+		}
```
ffalor: We could continue. Obv the API should behave correctly, but we have our limit, offset+limit, our exit condition is when the number of resources returned is 0.

ffalor: would be easier to implement the above if this method returned []cloudRiskModel instead of a set. It would also allow the data source to return the risks it did collect even if it fails partway through.

ffalor:
```Go
// for loop
		if res == nil || res.Payload == nil || len(res.Payload.Resources) == 0 {
			tflog.Debug(ctx, "[datasource] No more prevention policies to retrieve",
				map[string]interface{}{
					"total_retrieved": len(allPolicies),
				})
			break
		}



		// etc



		if res.Payload.Meta == nil || res.Payload.Meta.Pagination == nil ||
			res.Payload.Meta.Pagination.Offset == nil || res.Payload.Meta.Pagination.Total == nil {
			tflog.Warn(ctx, "Missing pagination metadata in API response, using offset+limit for next page",
				map[string]interface{}{
					"meta": res.Payload.Meta,
				})
			offset += limit
			continue
		}

		offset = int64(*res.Payload.Meta.Pagination.Offset)
		if offset >= *res.Payload.Meta.Pagination.Total {
			tflog.Info(ctx, "[datasource] Pagination complete",
				map[string]interface{}{
					"total_retrieved": len(allPolicies),
					"total_available": *res.Payload.Meta.Pagination.Total,
				})
			break
		}
```

PR #218, opened 2025-12-09
feat(cloud-azure-registration): call validate registration API in update event hub settings resource

cs-wayne-hsieh:


internal/fcs/cloud_azure_tenant_eventhub_settings.go
```diff
@@ -442,3 +443,38 @@ func (r *cloudAzureTenantEventhubSettingsResource) triggerHealthCheck(
...
+	var diags diag.Diagnostics
+	params := cloud_azure_registration.CloudRegistrationAzureValidateRegistrationParams{
+		Context:  ctx,
+		TenantID: tenantID,
+	}
+
+	_, err := r.client.CloudAzureRegistration.CloudRegistrationAzureValidateRegistration(&params)
+	if err != nil {
+		diags.AddWarning(
+			"Failed to validate registration. Please go to the Falcon console and trigger health check scan manually to reflect the latest state.",
```
ffalor: The wording here makes it seem like the act of triggering the health check failed, is that correct? Or should the wording be along the lines of "health check failed check the console for more details"

cs-wayne-hsieh: validate registration are the correct wording here because this is a validation call. When this occurs, the only thing that users can do is to trigger health check scan in Falcon console to get the current state.

PR #222, opened 2025-12-22
FCS AWS registration: Remove calls to cspmregistration

ashoksankarh-cs:
Remove calls to cspmregistration and call cloudreg APIs alone

### DIFF
--- a/internal/content_update_policy/content_update_policy_attachment.go
+++ b/internal/content_update_policy/content_update_policy_attachment.go
@@ -11,6 +11,7 @@ import (
 	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/framework/validators"
 	hostgroups "github.com/crowdstrike/terraform-provider-crowdstrike/internal/host_groups"
 	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/scopes"
+	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/tferrors"
 	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/utils"
 	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
@@ -23,7 +24,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
-	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
 var (
@@ -37,6 +37,18 @@ var (
 	attachmentResourceMarkdownDescription string         = "This resource allows managing the host groups attached to a content update policy. By default (when `exclusive` is true), this resource takes exclusive ownership over the host groups assigned to a content update policy. When `exclusive` is false, this resource only manages the specific host groups defined in the configuration. If you want to fully create or manage a content update policy please use the `content_update_policy` resource."
 	attachmentRequiredScopes              []scopes.Scope = apiScopesReadWrite
 )
+
+func newPolicyNotFoundError(policyID string) diag.ErrorDiagnostic {
+	return diag.NewErrorDiagnostic(
+		"Content Update Policy Not Found",
+		fmt.Sprintf(
+			"Content update policy with ID %q does not exist. "+
+				"This resource manages attachments to an existing policy and does not create a policy. "+
+				"Ensure the correct policy ID was provided or use the crowdstrike_content_update_policy resource to create a policy.",
+			policyID,
+		),
+	)
+}
 
 func NewContentUpdatePolicyAttachmentResource() resource.Resource {
 	return &contentUpdatePolicyAttachmentResource{}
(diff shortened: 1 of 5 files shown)


internal/fcs/cloud_aws_account_data_source.go
```diff
@@ -317,111 +269,72 @@ func (d *cloudAwsAccountsDataSource) Read(
 	if resp.Diagnostics.HasError() {
 		return
 	}
-	cspmAccounts, diags := d.getCSPMAccounts(
-		ctx,
-		data.AccountID.ValueString(),
-		data.OrganizationID.ValueString(),
-	)
+	// Get all cloud accounts without filtering first
+	cloudAccounts, diags := d.getCloudAccounts(ctx, []string{})
```
ffalor: What was the purpose behind removing the accountid/orgid filter if they provide one?

also `d.getCloudAccounts` does not handle pagination so there is no guarantee we get everything back. This is a bug can we fix it in this PR?

internal/fcs/cloud_aws_account.go
```diff
@@ -723,122 +716,230 @@ func (r *cloudAWSAccountResource) Create(
...
+// populateModelFromCloudAccount populates the Terraform model from CloudAwsRegistration response.
+func (r *cloudAWSAccountResource) populateModelFromCloudAccount(ctx context.Context, model *cloudAWSAccountModel, cloudAccount *models.DomainCloudAWSAccountV1) diag.Diagnostics {
 	var diags diag.Diagnostics
 
-	var targetOUs []string
-	if model.OrganizationID.ValueString() != "" {
-		diags.Append(model.TargetOUs.ElementsAs(ctx, &targetOUs, false)...)
+	// Set basic account information
+	model.AccountID = types.StringValue(cloudAccount.AccountID)
+	model.AccountType = types.StringValue(cloudAccount.AccountType)
```
ffalor: Need to add a check somewhere that verifies cloudAccount is not nil to prevent nil derefence issues.

internal/fcs/settings.go
```diff
@@ -55,6 +61,12 @@ func newSettingsConfig(ctx context.Context, settings interface{}, diags *diag.Di
 		LogIngestionSnsTopicArn:    types.StringNull(),
 		LogIngestionS3BucketPrefix: types.StringNull(),
 		LogIngestionKmsKeyArn:      types.StringNull(),
+
+		// Initialize DSPM and Vulnerability Scanning settings
+		DSPMRole:                           types.StringNull(),
```
ffalor: I would rename this to DSPMRoleName, since that is what we are calling it everywhere else in this codebase. The same for VulnerabilityScanningRole

internal/fcs/cloud_aws_account.go
```diff
@@ -723,122 +716,230 @@ func (r *cloudAWSAccountResource) Create(
...
-		AccountType:      model.AccountType.ValueString(),
-		DeploymentMethod: model.DeploymentMethod.ValueString(),
+	model.DspmRoleArn = types.StringValue(dspmRoleArn)
+	model.DspmRoleName = types.StringValue(dspmRoleName)
+	model.VulnerabilityScanningRoleArn = types.StringValue(vulnScanningRoleArn)
+	model.VulnerabilityScanningRoleName = types.StringValue(vulnScanningRoleName)
+
+	agentlessRoleName := resolveAgentlessScanningRoleNameV1(ctx, cloudAccount)
+	model.AgentlessScanningRoleName = types.StringValue(agentlessRoleName)
+
```
ffalor: there is no need to call newConfigSetting since we have the information here already
```suggestion
	switch {
	case model.DspmRoleName.ValueString() != "":
		model.AgentlessScanningRoleName = model.DspmRoleName
	case model.VulnerabilityScanningRoleName.ValueString() != "":
		model.AgentlessScanningRoleName = model.VulnerabilityScanningRoleName
	default:
		model.AgentlessScanningRoleName = types.StringNull()
	}

```

internal/fcs/cloud_aws_account.go
```diff
@@ -723,122 +716,230 @@ func (r *cloudAWSAccountResource) Create(
...
+// createCloudAccount creates a new Cloud AWS account from the resource model.
 func (r *cloudAWSAccountResource) createCloudAccount(
 	ctx context.Context,
 	model cloudAWSAccountModel,
 ) (*models.DomainCloudAWSAccountV1, diag.Diagnostics) {
 	var diags diag.Diagnostics
+	var targetOUs []string
+	if model.OrganizationID.ValueString() != "" {
+		diags.Append(model.TargetOUs.ElementsAs(ctx, &targetOUs, false)...)
+	}
```
ffalor: check if diags has an error and return it

internal/fcs/cloud_aws_account.go
```diff
@@ -951,16 +1048,13 @@ func (r *cloudAWSAccountResource) Read(
 		return
 	}
 
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
-	cspmAccount, diags := r.getCSPMAccount(ctx, state.AccountID.ValueString())
+	// Use CloudAwsRegistration only (replacing dual API calls)
```
ffalor:
```suggestion
```

ffalor (changes requested):

PR #223, opened 2026-01-05
feat(cloud_security_kac_policy): add new resource

mr-jungchoi:


internal/cloud_security/kac_policy_resource.go
```diff
@@ -0,0 +1,1253 @@
...
+			},
+			"host_groups": schema.SetAttribute{
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "Host Group ids to attach to the KAC policy.",
+			},
+			"rule_groups": schema.ListNestedAttribute{
+				Optional:    true,
+				Description: "A list of KAC policy rule groups in order of highest to lowest priority. Reordering the list will change rule group precedence. When reordering the list of rule groups to update precedence, the rule group names must match the state, otherwise the provider will consider it a new rule group, or an in place update.",
+				Validators: []validator.List{
```
ffalor: Similiar feedback to host_groups except we just need to require a size of atleast 1 to prevent `[]`

internal/cloud_security/kac_policy_resource.go
```diff
@@ -0,0 +1,1253 @@
...
+	if len(createResponse.Payload.Resources) == 0 {
+		resp.Diagnostics.AddError(
+			"Error creating KAC policy",
+			"No policy returned in create response",
+		)
+		return
+	}
+
+	policy := createResponse.Payload.Resources[0]
+	plan.ID = types.StringValue(*policy.ID)
```
ffalor: Set the ID of the policy we created because if we encounter an error before we run `resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)` the id will not be saved in state which will result in the policy being created, but untracked.

This is only needed when we have paths that can error and return before we do any `.Set`
```suggestion
	policy := createResponse.Payload.Resources[0]
	plan.ID = types.StringValue(*policy.ID)

    resp.Diagnostics.Append(
		resp.State.SetAttribute(ctx, path.Root("id"), plan.ID)...)
	if resp.Diagnostics.HasError() {
		return
	}
```

internal/cloud_security/kac_policy_resource.go
```diff
@@ -0,0 +1,1253 @@
...
+		updatedPolicy, hostGroupDiags := r.updateHostGroups(ctx, plan.ID.ValueString(), plan.HostGroups, basetypes.SetValue{})
+		resp.Diagnostics.Append(hostGroupDiags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
+		if updatedPolicy != nil {
+			policy = updatedPolicy
+		}
+	}
```
ffalor: host_groups will not be unknown here, and `utils.SetIDsToModify` can handle both null and unknown states. So it's safe to just call `r.updateHostGroups()` with no conditional as it will be able to handle null host groups and will not call the API unless needed, but if prefer to short-circuit then a simple len check can be used.
```suggestion
	if len(plan.HostGroups.Elements()) != 0 {
		updatedPolicy, hostGroupDiags := r.updateHostGroups(ctx, plan.ID.ValueString(), plan.HostGroups, basetypes.SetValue{})
		resp.Diagnostics.Append(hostGroupDiags...)
		if resp.Diagnostics.HasError() {
			return
		}

		if updatedPolicy != nil {
			policy = updatedPolicy
		}
	}
```

internal/cloud_security/kac_policy_resource.go
```diff
@@ -0,0 +1,1253 @@
...
+		if len(getResponse.Payload.Resources) == 0 {
+			diags.AddError(
+				"Error getting KAC policy",
+				"No policy returned in get policy response",
+			)
+			return nil, diags
+		}
+
+		apiKacPolicy = getResponse.Payload.Resources[0]
+	}
```
ffalor: Thoughts on making something like `r.getKACPolicy` that handles the all the error logic, nil checks etc and it can be used in both read and in here to prevent the need to maintain the error logic in two places etc.

internal/cloud_security/kac_policy_resource.go
```diff
@@ -0,0 +1,1253 @@
...
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
+				Description: "Whether the policy is enabled.",
+			},
+			"host_groups": schema.SetAttribute{
+				Optional:    true,
+				ElementType: types.StringType,
+				Description: "Host Group ids to attach to the KAC policy.",
+			},
```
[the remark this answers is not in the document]
mr-jungchoi: addressed

PR #236, opened 2026-01-12
feat: Add cloud security suppression rule

gpontejos-cs:


internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1251 @@
...
+						Validators: []validator.Set{
+							setvalidator.ValueStringsAre(
+								stringvalidator.OneOf("critical", "high", "medium", "informational"),
+							),
+						},
+					},
+				},
+			},
+			"scope_asset_filter": schema.SingleNestedAttribute{
+				MarkdownDescription: "Filter criteria for scope assets. Only necessary when `scope_type` is `asset_filter`.",
```
ffalor: scope_type doesn't look like a poassible attribute

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1251 @@
...
+					fwvalidators.StringNotWhitespace(),
+				},
+			},
+			"suppression_comment": schema.StringAttribute{
+				Description: "Comment for suppression. This will be attached to the Findings suppressed by this rule.",
+				Optional:    true,
+				Computed:    true,
+				Default:     stringdefault.StaticString(""),
+			},
+			"suppression_expiration_date": schema.StringAttribute{
```
ffalor: Whats the need to have a default and for this attribute to be computed?

Could we do something like this. `StringNotWhiteSpace()` requires the string to have a value or be null. `""`, `"   "` would fail validation. So this attribute will only ever be in two states.

Then you can use `StringValueToFramework` for the api response which will set `""` to null.

```suggestion
			"suppression_comment": schema.StringAttribute{
				Description: "Comment for suppression. This will be attached to the Findings suppressed by this rule.",
				Optional:    true,
			    Validators: []validator.String{
					validators.StringNotWhitespace(),
				},
			},
			"suppression_expiration_date": schema.StringAttribute{
```

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1202 @@
...
+						Optional:    true,
+						Validators: []validator.Set{
+							setvalidator.ValueStringsAre(
+								stringvalidator.LengthAtLeast(1),
+								fwvalidators.StringNotWhitespace(),
+							),
+						},
+					},
+					"cloud_providers": schema.SetAttribute{
+						MarkdownDescription: "Set of cloud providers. Examples: `aws`, `azure`, `gcp`. An Asset will match if it belongs to a cloud provider included in this set.",
```
ffalor: rule selection has `providers` that starts with this description `"Set of rule cloud providers. Examples: `AWS`, `Azure`, `GCP`, `OCI`. A rule will match if its cloud provider is included in this set."`

should these have similiar names?

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1202 @@
...
+		if diags.HasError() {
+			resp.Diagnostics.Append(diags...)
+		} else if expired {
+			resp.Diagnostics.AddAttributeError(
+				path.Root("expiration_date"),
+				"Expired Date",
+				"The expiration_date has already passed. If you are attempting to run a destroy operation, use 'terraform destroy -refresh=false' to skip the expiration check.",
+			)
+		}
+	}
```
ffalor: we check if plan is nil and return early so the comments/error message around destroy do not make sense.

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1202 @@
...
+				MarkdownDescription: "Expiration date for suppression. If defined, must be in RFC3339 format (e.g., `2025-08-11T10:00:00Z`). Once set, this field cannot be cleared. The suppression rule will still exist after expiration and can be reset by updating the expiration date.",
+				Optional:            true,
+				Validators: []validator.String{
+					fwvalidators.ValidateRFC3339(),
+					fwvalidators.StringNotWhitespace(),
+				},
+				PlanModifiers: []planmodifier.String{
+					fwmodifiers.PreventStringClearing("Suppression Expiration Date"),
+				},
+			},
```
ffalor: I wonder if the rfc3339 custom type would have any value here [common-custom-types](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom#common-custom-types) specifically around semantic equality

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1202 @@
...
+		return
+	}
+
+	if isRuleSelectionFilterEmpty(ruleSelectionFilter) {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("rule_selection_filter"),
+			"Empty Rule Selection Filter",
+			"When rule_selection_filter is defined, at least one filter criterion must be specified (ids, names, origins, providers, services, or severities).",
+		)
+	}
```
ffalor: same for asset scope filter, but  `rule_selection_filter` could be known, but an attribute under it could be unknown, so we would need to make sure the attributes are known too before we throw an error

ffalor (changes requested):

internal/framework/flex/time.go
```diff
@@ -0,0 +1,34 @@
...
+
+import "github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
+
+// RFC3339ValueToFramework converts a string to a Terraform framework timetypes.RFC3339.
+// An empty string returns a null timetypes.RFC3339.
+func RFC3339ValueToFramework[T ~string](v T) timetypes.RFC3339 {
+	if v == "" {
+		return timetypes.NewRFC3339Null()
+	}
+	rfc3339Value, _ := timetypes.NewRFC3339Value(string(v))
```
ffalor: if this errors it returns an unknown value we can't write unknown values to state I think we need to return diag and handle it.

internal/framework/validators/object_validators.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (v atLeastOneNonEmptyAttributeValidator) Description(_ context.Context) string {
+	return "At least one attribute must be non-empty (not null/unknown and containing actual values)"
+}
+
+func (v atLeastOneNonEmptyAttributeValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v atLeastOneNonEmptyAttributeValidator) ValidateObject(ctx context.Context, req validator.ObjectRequest, resp *validator.ObjectResponse) {
```
ffalor: nice

internal/framework/validators/string_validators.go
```diff
@@ -102,3 +102,20 @@ func SortField(validFields []string) validator.String {
...
+// Valid values: "2025-08-11T10:00:00Z", "" (empty string)
+// Invalid values: "2025-08-11", "10:00:00Z", "2025-08-11 10:00:00".
+func ValidateRFC3339() validator.String {
+	return stringvalidator.All(
+		stringvalidator.RegexMatches(
+			regexp.MustCompile(`^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z)?$`),
+			"must be in RFC3339 format (e.g., '2025-08-11T10:00:00Z') if defined",
+		),
+	)
+}
```
ffalor: doesnt look like this is used anywhere
```suggestion
```

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1082 @@
...
+	req resource.CreateRequest,
+	resp *resource.CreateResponse,
+) {
+	var plan cloudSecuritySuppressionRuleResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(r.validateExpirationDateFormat(plan.ExpirationDate)...)
```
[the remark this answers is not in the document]
ffalor: same goes for update if the above is true

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1082 @@
...
+			return nil, diags
+		}
+
+		body.RuleSelectionFilter, diags = ruleSelectionFilter.Expand(ctx)
+		if diags.HasError() {
+			return nil, diags
+		}
+
+		body.RuleSelectionType = ruleSelectionTypeFilter
+	}
```
ffalor: correct me if im wrong, but I think if i have a rule selection filter, and then delete that attribute, update will run, but the attribute will be null so this will be skipped so no changes are sent to the backend.

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1082 @@
...
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"type": schema.StringAttribute{
+				Description: "Type of suppression rule. Defaults to IOM.",
+				Required:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf(suppressionRuleSubdomainDefault),
+				},
+			},
```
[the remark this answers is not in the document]
ffalor: i updated this to One of similiar to your other enum description, but looks like it didnt make it before your apply

internal/cloud_security/suppression_rule_resource.go
```diff
@@ -0,0 +1,1160 @@
...
+			return
+		} else if expired {
+			resp.Diagnostics.AddAttributeError(
+				path.Root("expiration_date"),
+				"Expired Date",
+				"The expiration_date has already passed.",
+			)
+			return
+		}
+	}
```
ffalor: Modifyplan verifys expiration on updates so if we are here then the expiration is good, having the warning here would cause two warning messages one on read and one on update. I think it makes sense to keep the warning in read, thoughts?
```suggestion
```

PR #271, opened 2026-02-07
docs: improve plan-tf-resource skill workflow and add reference example

ffalor:
Improves the `plan-tf-resource` skill workflow with better package validation, clearer iteration process, and explicit guidelines on what to include/exclude in GitHub issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,7 @@ name: ci
 on:
   workflow_dispatch:
   pull_request:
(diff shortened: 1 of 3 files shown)


redhatrises (approved):

ffalor:
PR updated to remove path requirements form `ci.yaml` so the `generate` job can run and not block merge. Path check will be added back once permissions are fixed.

redhatrises (approved):

PR #300, opened 2026-02-26
feat: send Falcon Client ID to API in cloud azure tenant resource

flow-droid:


internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -151,6 +153,10 @@ func (r *cloudAzureTenantResource) Schema(
 				Required:            true,
 				MarkdownDescription: "The Azure Tenant ID to register into Falcon Cloud Security. If subscription_ids and management_group_ids are not provided, then all subscriptions in the tenant are targeted.",
 			},
+			"falcon_client_id": schema.StringAttribute{
+				MarkdownDescription: "Falcon API client ID.",
+				Optional:            true,
+			},
```
ffalor: checkout the google cloud registration, we can use the client id used to authenticate the provider.

ffalor (approved):

PR #306, opened 2026-02-28
feat: add data protection policy resource

### LINKED ISSUE #284: add `data_protection_policy` resource

GrantBirki:
This pull request adds the `crowdstrike_data_protection_policy` resource for Falcon Data Protection policy management in Terraform, including provider registration, sweeper support, examples, generated docs, and acceptance, unit, schema, and sweep test coverage. It also addresses review feedback by treating payload-level `404` responses as not found during `Read()` and by making legacy policy sweeps try both `win` and `mac` fallback platforms.

Resolves: https://github.com/CrowdStrike/terraform-provider-crowdstrike/issues/284


docs/resources/data_protection_policy.md
```diff
@@ -0,0 +1,165 @@
...
+  precedence    = 1
+
+  policy_properties {
+    classifications               = ["classification-id"]
+    enable_content_inspection     = true
+    min_confidence_level          = "medium"
+    max_file_size_to_inspect      = 10
+    max_file_size_to_inspect_unit = "MB"
+  }
+}
```
ffalor: The plugin framework has this on the blocks [documentation](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/blocks):
> Use [nested attribute types](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#nested-attribute-types) instead of block types for new schema implementations. Block support is mainly for migrating legacy SDK-based providers.

So we have not used block attributes in the provider yet. Question since I've never worked with them, are there any benefits to using a block attribute? 

Thanks again for your contribution, I'll take a closer look this week

ffalor:
This is being worked on, its just very large and I'm finding some api behavior that is just taking a little bit to validate. I've approved 2 of your others and will be merged once someone on my team is able to review my changes. the ml cert exclusion I'm also almost done going through.

omareltf:
@ffalor Why is it stuck? We really need it, together with the classification resource

ffalor:
Hey, due to the API there was just so many things that had to change, I ended up creating a separate pr #479 @GrantBirki thank you again for contributing. Also thank you and @omareltf for your patience sorry this took so long to get around to doing. 

As far as why it took so long, this is just a very large resource and the API does not play well with terraform so a lot was done to workaround that behavior to make the schema for the resource as simple as possible.

With that said the following settings can't be managed by the resource yet until the API is updated, i will be letting the team that owns the api endpoints know. When API support is added, I will add support to this resource.


## Console settings the public v2 API does not support

Six controls in the Falcon Data Protection policy console cannot be managed through the public v2 API, so `crowdstrike_data_protection_policy` cannot expose them. They fall into two groups by failure mode.

### 1. Write is accepted, then silently discarded

The API answers `200`, the value never takes effect, and the key is never returned by a `GET`.

| Console control | Wire field | Platform |
|---|---|---|
| Inspection > Printer inspection | `enable_print_monitor` | Windows |
| Inspection > Network inspection > Exclude applications | `network_inspection_exclude_list_v2` | Windows |
| Data upload via unsupported browsers (the action) | `unsupported_browsers_action` | Windows |

- `enable_print_monitor` and `network_inspection_exclude_list_v2` are also absent from gofalcon.
- `unsupported_browsers_action` is documented as a supported field, and the API reference uses it in its own example validation error, which does not reproduce. It is not the same field as `browsers_without_active_extension`, which is real and is managed.

### 2. Write is rejected, field is never returned

`PATCH .../policies/v2?platform_name=mac` answers `400 "<field> is not a field of Data Protection Mac policy"`, and a Mac `GET` returns 36 `policy_properties` keys with no `evidence_*` key, even while the setting is enabled in the console.

| Console control | Wire field | Platform |
|---|---|---|
| Forensics > Evidence storage and retrieval | `evidence_download_enabled` | Mac |
| Evidence storage limits > Max size (GiB) | `evidence_storage_max_size` | Mac |
| Evidence storage limits > Max free space to consume (%) | `evidence_storage_free_disk_perc` | Mac |

This is the only bidirectional gap, and it is why evidence storage is Windows-only in the resource. Five `evidence_*` wire fields are rejected on Mac; the three above are the ones with Mac console controls. The other two are `evidence_encrypted_enabled`, exposed as the Windows-only `end_user_encryption_activity`, and `evidence_duplication_enabled_default`, which has no console control on either platform and is not exposed at all. The console does not use the public API for any of this: it PATCHes an internal v1 endpoint with a PascalCase `inputs` map, and sending the console's own spelling `EvidenceDownloadEnabled` to v2 returns `200` and changes nothing.

PR #311, opened 2026-03-05
feat: Migrate cloud_security_custom_rule to cloud_security_iom_custom_rule

gpontejos-cs:


ffalor (commented):
Two small bugs in `shared.go`:

**Line 244**: `NewDiagnosticFromAPIError` is called twice — the inner `diag` shadows the outer one with the same call. The inner declaration should be removed:
```go
if diag != nil {
    // diag := tferrors.NewDiagnosticFromAPIError(tferrors.Delete, err, scopes)  // remove this line
    if diag.Summary() == tferrors.NotFoundErrorSummary {
```

**Line 231**: `updateCloudPolicyRule` uses `tferrors.Read` instead of `tferrors.Update` for the payload error diagnostic, which would produce a misleading error message:
```go
diags.Append(tferrors.NewOperationError(tferrors.Update, err))  // should be Update, not Read
```

PR #317, opened 2026-03-11
feat: add host_group data source

yoclaire:
Adds a `crowdstrike_host_group` data source for looking up a host group by name or ID.

Co-authored with Claude Code, let me know if there are any concerns with this!


ffalor (approved):
Thank you for contributing! I'm sorry it took so long to get to this.

redhatrises (approved):

Issue #175, opened 2025-11-07
Add data source for NG-SIEM Rules


vova-bolotov:
## Description
When creating custom [SIEM correlation rules](https://falcon.crowdstrike.com/xdr/rules), the rules are automatically assigned to the admin user who created them. During Fal.Con Europe 2025, I learned that if this admin account is later deleted, all rules owned by that admin are also deleted, which can lead to unintended data loss and operational disruptions.

To prevent this, it would be beneficial to allow rules to be created and managed via Terraform, ensuring that ownership is not tied to an individual user.

## API Scopes Required

- Correlation Rules Admin `RIGHT`
- Correlation Rules `READ` `RIGHT`

ffalor:
It does look like there is an API for this. Everything you would configure that create rule workflow looks like is one api call.

@cfitzgerald-pd could you point me to another provider that has a resource like this?

petergs:
@ffalor I'd also love to see this as it's a serious blocker for us to consider using NG-SIEM as a replacement for our existing SIEM. The Terraform provider for both [Sumo Logic Cloud SIEM](https://registry.terraform.io/providers/sumologic/sumologic/latest/docs/resources/cse_match_rule) and [Datadog Cloud SIEM ](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/security_monitoring_rule) have resources for threat detection rules.

PR #322, opened 2026-03-18
feat(correlation_rules): create resource & data source

RainerEndres:
Hello, 

based on issue #175 I added 
-  the `crowdstrike_correlation_rule` resource
-  the `crowdstrike_correlation_rules` data source

Since I could not find to extensive documentation on some parts of the API some interactions are based on assumptions. Feel free to contact me early, ideally with precise documentation, such that I can resolve such issues quickly. 

There is an issue with the gofalcon library, in that for the patch API `omitempty` is set for many members. 
This causes a forced recreate for some members, since otherwise setting a value considered "empty"  would not cause any change. 
If I find the time and a proper way to fix this in the gofalcon library, I'll do so and create another PR, once changes to gofalcon are merged. 

Thank you
- Rainer


Jperkinscarecom:
+1, blocking on this for our IaC adoption

ffalor (approved):

redhatrises (approved):

PR #367, opened 2026-05-12
feat(cloud_azure_tenant): add agentless_scanning_subscription_ids attribute

zango88:
## Summary                                                                                                                                                                                                                            
                                                            
  - Add optional `agentless_scanning_subscription_ids` (Set of String) attribute to `crowdstrike_cloud_azure_tenant`                                                                                                                    
  - Send subscription IDs as `additional_features` to the CrowdStrike API on Create and Update
  - Fix `dspm.enabled` reading from `additional_features` when subscription-scoped DSPM is configured                                                                                                                                   
                                                                                                                                                                                                                                        
  ## Context                                                                                                                                                                                                                            
                                                                                                                                                                                                                                        
  When using tenant-wide registration with limited DSPM subscriptions (`agentless_scanning_locations_per_subscription`), the API stores DSPM in `additional_features` rather than `products.features`. This caused:                     
  1. Provider couldn't pass subscription IDs to the API
  2. `dspm.enabled` was incorrectly `false` after apply → "inconsistent result" error                                                                                                                                                   
                                                                                                                                                                                                                                        
  ## Changes                                                                                                                                                                                                                            
                                                                                                                                                                                                                                        
  | File | Change |                                                                                                                                                                                                                     
  |------|--------|                                         
  | `internal/fcs/cloud_azure_tenant_resource.go` | Add field to model, schema, Create, Update, and `wrap()` |                                                                                                                          
  | `docs/resources/cloud_azure_tenant.md` | Auto-generated |                                                                                                                                                                           
                                                                                                                                                                                                                                        
  ## Backward Compatibility                                                                                                                                                                                                             
                                                                                                                                                                                                                                        
  Attribute is optional — omitting it sends no `additional_features`. Existing configs unchanged.                                                                                                                                       
                                                            
  ## Test plan                                                                                                                                                                                                                      
                                                            
  - [V] `terraform apply` with `agentless_scanning_subscription_ids` set (tenant-wide scenario)                                                                                                                                         
  - [V] `terraform apply` without the attribute (regression)
  - [V] Verify `dspm.enabled` reads correctly from `additional_features` response                                                                                                                                                       
  - [V] Verify `terraform destroy` cleans up properly

### DIFF
--- a/internal/fcs/cloud_azure_tenant_resource.go
+++ b/internal/fcs/cloud_azure_tenant_resource.go
@@ -322,11 +322,14 @@ func (m *cloudAzureTenantModel) wrap(
 		}
 	}
 
-	var agentlessSubIdsSlice []string
 	for _, af := range registration.AdditionalFeatures {
 		if af != nil && af.Feature != nil && *af.Feature == "dspm" {
-			hasDSPM = true
-			agentlessSubIdsSlice = af.SubscriptionIds
+			if !m.AgentlessScanningSubscriptionIds.IsNull() && len(af.SubscriptionIds) > 0 {
+				hasDSPM = true
+				agentlessSubIds, d := flex.FlattenStringValueSet(ctx, af.SubscriptionIds)
+				diags.Append(d...)
+				m.AgentlessScanningSubscriptionIds = agentlessSubIds
+			}
 			break
 		}
 	}
@@ -340,12 +343,6 @@ func (m *cloudAzureTenantModel) wrap(
 	dspmObj, d := dspm.ToObject(ctx)
 	diags.Append(d...)
 	m.DSPM = dspmObj
-
-	if !m.AgentlessScanningSubscriptionIds.IsNull() && len(agentlessSubIdsSlice) > 0 {
-		agentlessSubIds, d := flex.FlattenStringValueSet(ctx, agentlessSubIdsSlice)
-		diags.Append(d...)
-		m.AgentlessScanningSubscriptionIds = agentlessSubIds
-	}
 
 	return diags
 }
--- a/internal/fcs/cloud_azure_tenant_resource.go
+++ b/internal/fcs/cloud_azure_tenant_resource.go
@@ -82,20 +82,21 @@ func (r *cloudAzureTenantResource) Metadata(
 }
 
 type cloudAzureTenantModel struct {
-	AccountType                 types.String `tfsdk:"account_type"`
-	AppRegistrationId           types.String `tfsdk:"cs_azure_client_id"`
-	CsInfraRegion               types.String `tfsdk:"cs_infra_location"`
-	CsInfraSubscriptionId       types.String `tfsdk:"cs_infra_subscription_id"`
-	Environment                 types.String `tfsdk:"environment"`
-	SubscriptionIds             types.Set    `tfsdk:"subscription_ids"`
-	ManagementGroupIds          types.Set    `tfsdk:"management_group_ids"`
-	MicrosoftGraphPermissionIds types.Set    `tfsdk:"microsoft_graph_permission_ids"`
-	ResourceNamePrefix          types.String `tfsdk:"resource_name_prefix"`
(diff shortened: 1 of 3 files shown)


internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -693,6 +722,17 @@ func (r *cloudAzureTenantResource) updateRegistration(
...
+	if len(agentlessSubIds) > 0 {
+		params.Body.Resource.AdditionalFeatures = []*models.AzureAdditionalFeature{
+			{
+				Feature:         utils.Addr("dspm"),
+				Product:         utils.Addr("cspm"),
+				SubscriptionIds: agentlessSubIds,
+			},
+		}
+	}
+
```
ffalor: we pass nothing on update if the number of ids is 0, will this cause the backend to reset the ids if configured? That would be the expected behavior here

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -257,6 +258,11 @@ func (r *cloudAzureTenantResource) Schema(
 				Optional:            true,
 				MarkdownDescription: "Tags applied to managed resources. This does not effect the registration of the tenant. It will be used if you generate new .tfvars from the UI.",
 			},
+			"agentless_scanning_subscription_ids": schema.SetAttribute{
```
ffalor, line 303: should this be in the `dspm` block?

ffalor, line 303:
```go
				Validators: []validator.Set{
					setvalidator.SizeAtLeast(1),
					setvalidator.ValueStringsAre(
						validators.StringNotWhitespace(),
					),
				},
```

Add these validators, this will prevent `agentless_scanning_subscription_ids = []` which means `agentless_scanning_subscription_ids` will either be null or have 1 or more values. This simplifies the wrap logic so we can treat an empty list as null.

This pattern came after this resource was created so the current sets don't do this, and the only reason they havent been changed is because it would be a breaking change.

ffalor (changes requested):

internal/fcs/cloud_azure_tenant_resource.go
AdarShakedCrowdstrike: we will add vul scanning soon and this one is shared, although we will set them for each feature from API prespective

internal/fcs/cloud_azure_tenant_resource.go
zango88: Yes — when we don't pass additional_features on update, the backend's deleteScopedRegistrations deletes all the related rows (no filter applied when additionalFeatureSubIDs is empty). So the IDs are cleared as expected

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -502,14 +510,35 @@ func (r cloudAzureTenantResource) ModifyPlan(
...
 
-	var state, plan cloudAzureTenantModel
-	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	var plan cloudAzureTenantModel
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !plan.AgentlessScanningSubscriptionIds.IsNull() {
```
ffalor: We will also need to check if it is unknown, before doing validation. Without checking that you could throw an error because .ValueBool returns false when the value is unknown. ValidateConfig runs multiple times and since this is not a computed attributed we are guranteed to sooner or later have a known value.

An example of how this bug works is documented here: https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/a0348df12fae6b2692c2b853c168dc36bb09732d/internal/sensor_update_policy/default_sensor_update_policy_resource_test.go#L149

ffalor (dismissed):

ffalor (approved):

redhatrises (approved):

PR #401, opened 2026-05-21
fix(ioa_rule_group): accept windows ruletype_id 5 on read

### LINKED ISSUE #400: `crowdstrike_ioa_rule_group` read fails on Windows rules with `ruletype_id=5` ("Unknown rule type ID")
## Summary

The `crowdstrike_ioa_rule_group` resource in provider v0.0.72 errors during `terraform plan`/`apply`/`import` for any IOA rule group whose Falcon-side rules carry `ruletype_id = "5"` on platform `Windows`. Falcon returns id `5` for some pre-existing Windows Process Creation rules (the provider only writes id `1` for that type), and the read-time lookup has no way to express that alias. This blocks managing/importing legitimate Windows Process Creation rule groups that were created outside the provider or under an older API.

## Reproduction

- Provider: `CrowdStrike/crowdstrike` v0.0.72 (also reproduces against `main` @ `da08fd7`)
- A `crowdstrike_ioa_rule_group` resource for a Windows rule group `<group-id>` containing 2 rules of type "Process Creation" where the Falcon API returns `ruletype_id = "5"` for each.
- Run `terraform plan` (or `terraform import`).

Error:

```
Error: Unknown rule type ID
Rule type ID "5" for platform "Windows" is not recognized by this provider version.
```

## Root cause

`internal/ioa_rule_group/ioa_rule_group_resource.go` lines 66-85 declares a flat name→ID map:

```go
var ruleTypeIDMap = map[string]map[string]string{
    "Windows": {
        "Process Creation":   "1",
        "File Creation":      "2",
        "Network Connection": "9",
        "Domain Name":        "11",
    },
    "Linux": { ... },
    "Mac":   { "Process Creation": "5", ... },
}

var ruleTypeNameMap = invertNestedMap(ruleTypeIDMap)
```

Because the source map is name→ID, the inverted ID→name map can only carry one API ID per type. Falcon's actual Windows ruletype space includes `1` AND `5` for "Process Creation"; the provider only knows about `1`. On read (lines 635-646):

```go
ruleTypeName := ""
if apiRule.RuletypeID != nil {
    if nameMap, ok := ruleTypeNameMap[platform]; ok {
        if name, found := nameMap[*apiRule.RuletypeID]; found {
            ruleTypeName = name
        } else {
            diags.AddError(
                "Unknown rule type ID",
                fmt.Sprintf("Rule type ID %q for platform %q is not recognized by this provider version.", *apiRule.RuletypeID, platform),
            )
            return types.ListNull(types.ObjectType{AttrTypes: ruleAttrTypes}), diags
        }
    }
}
```

`"5"` for Windows misses the inverted map and the resource bails.

## Suggested fix

Keep `ruleTypeIDMap` as the canonical write-side map (one Terraform `type` → one API id used when creating rules), and add a small read-side alias map listing the additional API ids that should fold into the same Terraform `type`. On read, check the inverted map first, then the alias map, then error:

```go
// Additional API rule-type IDs observed in Falcon that should map to the same
// Terraform `type` enum on read. Write path is unaffected (uses ruleTypeIDMap).
var ruleTypeReadAliasMap = map[string]map[string]string{
    "Windows": {
        "5": "Process Creation",
    },
}

// ... in the read flow:
if name, found := nameMap[*apiRule.RuletypeID]; found {
    ruleTypeName = name
} else if aliasMap, ok := ruleTypeReadAliasMap[platform]; ok {
    if name, found := aliasMap[*apiRule.RuletypeID]; found {
        ruleTypeName = name
    }
}
if ruleTypeName == "" {
    diags.AddError("Unknown rule type ID", ...)
    return ...
}
```

A more general shape if more aliases turn up: switch the write-side to `map[string]map[string][]string` (Terraform type → list of API ids, first is canonical for writes) and derive both maps from it.

## Notes

- The user-facing schema validator at `ioa_rule_group_resource.go:502` already pins `type` to `OneOf("Process Creation", "File Creation", "Network Connection", "Domain Name")`, so the alias map doesn't widen the write contract — users still can't request the `5` variant explicitly; it only round-trips correctly.
- Mac already maps `Process Creation` → `5` for a different platform, so collapsing multiple API ids per platform→Terraform-type pair is consistent with the existing model.
- It would be worth Falcon-side auditing whether Linux or Mac have analogous undercovered ruletype ids; this patch only fixes the observed Windows/5 case.

yoclaire:
Closes #400.

Adds a read-side alias map (`ruleTypeReadAliasMap`) so multiple Falcon API rule-type IDs can collapse to one Terraform `type` enum value on read. Concrete case: Windows Process Creation rules created outside the provider can carry `ruletype_id = "5"` (the provider writes `"1"`); the read path previously errored with "Unknown rule type ID" on `terraform plan` / `import`.

Write path is unchanged — `ruleTypeIDMap` still owns one canonical API ID per Terraform `type`, and the schema `OneOf` validator at `ioa_rule_group_resource.go:502` still pins users to the four canonical type names. The alias map only affects round-trip on read/import/refresh.

Tested with `go test ./...` (all packages pass). Did not run `make acctest` — requires `FALCON_CLIENT_ID` / `FALCON_CLIENT_SECRET`.

See #400 for the original error text, repro, and a more general "type → list of API ids" shape that could replace this if more aliases turn up across platforms. This PR keeps the change minimal and scoped to the observed Windows/5 case.

### DIFF
--- a/internal/ioa_rule_group/ioa_rule_group_resource.go
+++ b/internal/ioa_rule_group/ioa_rule_group_resource.go
@@ -86,6 +86,21 @@ var ruleTypeIDMap = map[string]map[string]string{
 
 var ruleTypeNameMap = invertNestedMap(ruleTypeIDMap)
 
+// ruleTypeReadAliasMap holds additional Falcon API rule-type IDs that should
+// fold into the same Terraform `type` enum value on read. The write path is
+// unaffected: rules created by the provider use the canonical IDs in
+// ruleTypeIDMap. Some Falcon-side rules (typically created outside the
+// provider or under earlier API versions) carry these alias IDs, and the
+// resource must accept them on read/import without erroring.
+var ruleTypeReadAliasMap = map[string]map[string]string{
+	"Windows": {
+		// Some Windows Process Creation rules return id "5" (the canonical
+		// write-side id is "1"); see CrowdStrike/terraform-provider-crowdstrike
+		// upstream issue for details.
+		"5": "Process Creation",
+	},
+}
+
 var dispositionMap = map[string]int32{
 	"Monitor":      10,
 	"Detect":       20,
@@ -637,7 +652,12 @@ func wrapRules(
 			if nameMap, ok := ruleTypeNameMap[platform]; ok {
 				if name, found := nameMap[*apiRule.RuletypeID]; found {
 					ruleTypeName = name
-				} else {
+				} else if aliasMap, ok := ruleTypeReadAliasMap[platform]; ok {
+					if name, found := aliasMap[*apiRule.RuletypeID]; found {
+						ruleTypeName = name
+					}
+				}
+				if ruleTypeName == "" {
 					diags.AddError(
 						"Unknown rule type ID",
 						fmt.Sprintf("Rule type ID %q for platform %q is not recognized by this provider version.", *apiRule.RuletypeID, platform),
--- a/internal/ioa_rule_group/ioa_rule_group_test.go
+++ b/internal/ioa_rule_group/ioa_rule_group_test.go
@@ -162,3 +162,73 @@ func TestConvertIOARuleGroupToList(t *testing.T) {
 		})
 	}
 }
+
+func TestWrapRulesRuleTypeIDLookup(t *testing.T) {
+	int32Ptr := func(v int32) *int32 { return &v }
+
+	tests := []struct {
+		name       string
+		platform   string
+		ruletypeID string
+		wantType   string
+		wantErr    bool
+	}{
+		{
+			name:       "windows canonical id 1 resolves to Process Creation",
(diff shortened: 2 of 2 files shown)


ffalor:
Thanks for catching this and putting up a fix, @yoclaire! 

We ended up taking a broader approach in #451. Instead of aliasing `ruletype_id=5`, we switched the read path to resolve the type from the API's `ruletype_name` field directly. While investigating we found the same tenant also had a rule with `ruletype_id=3` that a `5`-only alias wouldn't have covered, and since we can't know which other legacy ids are out there, reading by name resolves all of them in one shot.

Going to close this in favor of #451, but the report and your investigation were what pointed us at the root cause. Really appreciate it!

PR #406, opened 2026-06-10
fix(compliance_custom_framework): use UseNonNullStateForUnknown for control id

bmwambazi-cs:
When a crowdstrike_cloud_compliance_custom_framework resource already exists in state with sections and controls, adding a new control fails with `"Provider produced inconsistent result after apply"`. The control's id is null in the plan but receives a UUID after apply.

### Root Cause
The controls[*].id attribute uses UseStateForUnknown(), which copies the prior state value into the plan during updates. Since the resource already exists (making this an Update, not a Create), the modifier fires, but for a new control that didn't exist in prior state, the state value at that path is null. This overwrites the expected unknown with null. Terraform accepts unknown→value transitions but rejects null→value transitions.

###  How to Replicate
  1. Apply a framework with at least one section and one control (so the resource exists in state)
  2. Add a new control to any section (existing or new)
  3. Run terraform apply, fails with "inconsistent result after apply" on the new control's id

  ### Fix
Replace UseStateForUnknown() with UseNonNullStateForUnknown() on the control id attribute. This variant skips copying when the state value is null (new entry), leaving the plan as unknown so Terraform accepts the UUID assigned after apply.

Fixes: SUPPORT-50542

### DIFF
--- a/internal/cloud_compliance/custom_framework_resource_test.go
+++ b/internal/cloud_compliance/custom_framework_resource_test.go
@@ -823,6 +823,15 @@ func TestAccCloudComplianceCustomFrameworkResource_ComprehensiveCRUD(t *testing.
 							customFrameworkResourceName,
 							plancheck.ResourceActionUpdate,
 						),
+						// Verify new control IDs are unknown (not null) in plan - reproduces SUPPORT-50542
+						plancheck.ExpectUnknownValue(
+							customFrameworkResourceName,
+							tfjsonpath.New("sections").AtMapKey("section-1").AtMapKey("controls").AtMapKey("control-1.1").AtMapKey("id"),
+						),
+						plancheck.ExpectUnknownValue(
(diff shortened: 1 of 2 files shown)


internal/cloud_compliance/custom_framework_resource_test.go
```diff
@@ -823,6 +823,15 @@ func TestAccCloudComplianceCustomFrameworkResource_ComprehensiveCRUD(t *testing.
 							customFrameworkResourceName,
 							plancheck.ResourceActionUpdate,
 						),
+						// Verify new control IDs are unknown (not null) in plan - reproduces SUPPORT-50542
```
mr-jungchoi:
```suggestion
						// Verify new control IDs are unknown (not null) in plan
```

mr-jungchoi (approved):

PR #411, opened 2026-06-18
feat: add vulnerability_scanning attribute to azure tenant resource -…

JessicaKane213:
… CSPG-87057

Add vulnerability_scanning nested attribute (with enabled bool) to crowdstrike_cloud_azure_tenant resource, matching the pattern used by dspm. Sends "vulnerability_scanning" feature to the CSPM product config. Updates ModifyPlan to accept either dspm or vulnerability_scanning for agentless_scanning_subscription_ids validation.

### DIFF
--- a/internal/fcs/cloud_azure_tenant_resource.go
+++ b/internal/fcs/cloud_azure_tenant_resource.go
@@ -98,6 +98,7 @@ type cloudAzureTenantModel struct {
 	TenantId                         types.String `tfsdk:"tenant_id"`
 	RealtimeVisibility               types.Object `tfsdk:"realtime_visibility"`
 	DSPM                             types.Object `tfsdk:"dspm"`
+	VulnerabilityScanning            types.Object `tfsdk:"vulnerability_scanning"`
 	AgentlessScanningSubscriptionIds types.Set    `tfsdk:"agentless_scanning_subscription_ids"`
 }
 
@@ -137,6 +138,24 @@ func (r *dspmModel) FromObject(ctx context.Context, obj types.Object) diag.Diagn
 	return obj.As(ctx, r, basetypes.ObjectAsOptions{})
 }
 
+type vulnerabilityScanningModel struct {
+	Enabled types.Bool `tfsdk:"enabled"`
+}
+
+func (r *vulnerabilityScanningModel) AttributeTypes() map[string]attr.Type {
+	return map[string]attr.Type{
+		"enabled": types.BoolType,
+	}
+}
+
+func (r *vulnerabilityScanningModel) ToObject(ctx context.Context) (types.Object, diag.Diagnostics) {
+	return types.ObjectValueFrom(ctx, r.AttributeTypes(), r)
+}
+
+func (r *vulnerabilityScanningModel) FromObject(ctx context.Context, obj types.Object) diag.Diagnostics {
+	return obj.As(ctx, r, basetypes.ObjectAsOptions{})
+}
+
 func buildAzureProductConfig(
 	ctx context.Context,
 	data *cloudAzureTenantModel,
@@ -159,6 +178,17 @@ func buildAzureProductConfig(
 	if diags.HasError() {
 		return features, additionalFeatures
 	}
+
+	var vulnScanning vulnerabilityScanningModel
+	diags.Append(vulnScanning.FromObject(ctx, data.VulnerabilityScanning)...)
+	if diags.HasError() {
+		return features, additionalFeatures
+	}
+
+	if vulnScanning.Enabled.ValueBool() {
+		features = append(features, "vulnerability_scanning")
+	}
+
 	if !dspm.Enabled.ValueBool() {
 		return features, additionalFeatures
 	}
@@ -276,6 +306,27 @@ func (r *cloudAzureTenantResource) Schema(
 					),
 				),
 			},
+			"vulnerability_scanning": schema.SingleNestedAttribute{
+				Required: false,
+				Optional: true,
+				Computed: true,
+				Attributes: map[string]schema.Attribute{
+					"enabled": schema.BoolAttribute{
+						Required:    true,
+						Description: "Enable Vulnerability Scanning",
+					},
+				},
+				Default: objectdefault.StaticValue(
+					types.ObjectValueMust(
+						map[string]attr.Type{
+							"enabled": types.BoolType,
+						},
+						map[string]attr.Value{
+							"enabled": types.BoolValue(false),
+						},
+					),
+				),
+			},
 			"resource_name_prefix": schema.StringAttribute{
 				MarkdownDescription: "The prefix added to resources created during onboarding. It will be used if you generate new .tfvars from the UI.",
 				Optional:            true,
@@ -368,6 +419,7 @@ func (m *cloudAzureTenantModel) wrap(
 
 	hasIOA := false
 	hasDSPM := false
+	hasVulnScanning := false
 	for _, product := range registration.Products {
 		if *product.Product == "cspm" {
 			for _, feature := range product.Features {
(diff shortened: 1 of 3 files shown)


internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -159,21 +178,59 @@ func buildAzureProductConfig(
...
+
+	if dspm.Enabled.ValueBool() {
+		features = append(features, "dspm")
+	}
+
+	if !dspm.Enabled.ValueBool() && !vulnScanning.Enabled.ValueBool() {
 		return features, additionalFeatures
 	}
 
 	subs := flex.ExpandSetAs[string](ctx, data.AgentlessScanningSubscriptionIds, diags)
```
flow-droid:
```suggestion
	agentlessSubscriptionIDs := flex.ExpandSetAs[string](ctx, data.AgentlessScanningSubscriptionIds, diags)
```

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -159,21 +178,59 @@ func buildAzureProductConfig(
...
-		SubscriptionIds: subs,
-	}}
+	// Send subscription IDs as additional features for whichever scanning features are enabled
+	if dspm.Enabled.ValueBool() {
+		additionalFeatures = append(additionalFeatures, &models.AzureAdditionalFeature{
+			Feature:         utils.Addr("dspm"),
+			Product:         utils.Addr("cspm"),
+			SubscriptionIds: subs,
+		})
+		// Remove "dspm" from top-level features since it's now in additionalFeatures
```
flow-droid: I think we should not add it in the first place. I'm thinking 
```
if dspm.Enabled.ValueBool() {
if len(subs) == 0 {
// add as feature
} else {
// add as additional feature
}
}
```

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -159,21 +178,59 @@ func buildAzureProductConfig(
...
+
+	if !dspm.Enabled.ValueBool() && !vulnScanning.Enabled.ValueBool() {
 		return features, additionalFeatures
 	}
 
 	subs := flex.ExpandSetAs[string](ctx, data.AgentlessScanningSubscriptionIds, diags)
 	if len(subs) == 0 {
-		features = append(features, "dspm")
 		return features, additionalFeatures
 	}
```
flow-droid: I think I would remove such early returns, as those can be confusing if non-agentless feature is introduced

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -376,21 +455,28 @@ func (m *cloudAzureTenantModel) wrap(
...
-		if af != nil && af.Feature != nil && *af.Feature == "dspm" {
-			agentlessSubIds = af.SubscriptionIds
-			break
+		if af != nil && af.Feature != nil {
+			switch *af.Feature {
+			case "dspm":
+				agentlessSubIds = af.SubscriptionIds
+				hasDSPM = true
+			case "vulnerability_scanning":
+				if len(agentlessSubIds) == 0 {
```
flow-droid: why do we have this case? if it's additional feature it should have sub IDs

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -376,21 +455,28 @@ func (m *cloudAzureTenantModel) wrap(
...
+					agentlessSubIds = af.SubscriptionIds
+				}
+				hasVulnScanning = true
+			}
 		}
 	}
-	if len(agentlessSubIds) > 0 {
-		hasDSPM = true
-	}
 	agentlessSet, d := flex.FlattenStringValueSet(ctx, agentlessSubIds)
```
flow-droid: nit: agentlessSubscriptionsSet

internal/fcs/cloud_azure_tenant_resource.go
flow-droid, line 1: please add acceptance tests for different vuln scanning cases

internal/fcs/cloud_azure_tenant_resource.go
```diff
@@ -159,21 +178,59 @@ func buildAzureProductConfig(
...
-	}}
+	// Send subscription IDs as additional features for whichever scanning features are enabled
+	if dspm.Enabled.ValueBool() {
+		additionalFeatures = append(additionalFeatures, &models.AzureAdditionalFeature{
+			Feature:         utils.Addr("dspm"),
+			Product:         utils.Addr("cspm"),
+			SubscriptionIds: subs,
+		})
+		// Remove "dspm" from top-level features since it's now in additionalFeatures
+		for i, f := range features {
```
AdarShakedCrowdstrike: try using 
features = slices.DeleteFunc(features, func(f string) bool {
    return f == "dspm"
})

why this deletion is needed?

internal/fcs/cloud_azure_tenant_resource.go
AdarShakedCrowdstrike: +1

internal/fcs/cloud_azure_tenant_resource.go
JessicaKane213: Already removed - lmk your thoughts

flow-droid (dismissed):

internal/fcs/cloud_azure_tenant_resource_test.go
```diff
@@ -509,11 +509,11 @@ func TestAccCloudAzureTenant_agentless_UnknownDSPMEnabled(t *testing.T) {
 			},
 			{
 				Config:      testAccCloudAzureTenantConfig_agentlessSubIdsUnknownDSPM(tenantID, subID1, false),
-				ExpectError: regexp.MustCompile("agentless_scanning_subscription_ids requires dspm to be enabled"),
+				ExpectError: regexp.MustCompile("agentless_scanning_subscription_ids requires dspm or vulnerability_scanning to be enabled"),
```
AdarShakedCrowdstrike: add test for UnknownVulnerabilityScanningEnabled

AdarShakedCrowdstrike (dismissed):

AdarShakedCrowdstrike (dismissed):

ffalor:
The 6 `ExpectError` regexes in `internal/fcs/cloud_azure_tenant_resource_test.go` no longer match the diagnostic, so the modified/new acceptance tests fail at plan time.

The lengthened validation message now wraps across two lines when Terraform renders it:

```
agentless_scanning_subscription_ids requires dspm or vulnerability_scanning
to be enabled.
```

The regex has a literal space between `vulnerability_scanning` and `to`, but the rendered output has a newline + indentation there, so it never matches. Running it:

```
$ TF_ACC=1 go test ./internal/fcs/ -run TestAccCloudAzureTenant_agentlessSubIdsRequiresDSPMOrVulnScanning -v
    cloud_azure_tenant_resource_test.go:480: Step 1/1, expected an error with pattern, no match on: ...
        agentless_scanning_subscription_ids requires dspm or vulnerability_scanning
        to be enabled.
--- FAIL: TestAccCloudAzureTenant_agentlessSubIdsRequiresDSPMOrVulnScanning
```

This affects 6 `ExpectError` lines (486, 512, 516, 542, 568, 594), across `_agentlessSubIdsRequiresDSPMOrVulnScanning`, `_agentless_UnknownDSPMEnabled`, `_agentless_UnknownSubIds`, `_agentless_UnknownDSPMObject`, and the new `_agentless_UnknownVulnerabilityScanningEnabled`. The validation logic itself is correct; only the test assertions are broken by the line wrap.

Suggested fix: switch to a raw-string regex and relax the wrap point with `\s+`:

```go
ExpectError: regexp.MustCompile(`agentless_scanning_subscription_ids requires dspm or vulnerability_scanning\s+to be enabled`),
```

Verified locally: with this change the full `internal/fcs` acceptance suite passes (all previously-failing Azure tests plus the new vulnerability_scanning coverage).

ffalor (changes requested):
Once the acc tests are updated we can merge

ffalor (approved):

PR #410, opened 2026-06-18
feat(ngsiem): add crowdstrike_ngsiem_data_connection resource

ahay-8am:
## Summary

Adds a new resource, `crowdstrike_ngsiem_data_connection`, that manages CrowdStrike Falcon **Next-Gen SIEM data connections** (data connectors) as code — for example an AWS CloudTrail/S3 log source ingested via SQS notifications. Today these can only be created through the Falcon console wizard / AWS Marketplace automation; this resource closes that IaC gap.

It wraps the existing `gofalcon` `ngsiem` client's external data-connection operations:
- `ExternalCreateDataConnection` (Create)
- `ExternalGetDataConnectionByID` (Read)
- `ExternalUpdateDataConnection` (Update)
- `ExternalDeleteDataConnection` (Delete)

No new SDK dependency — `gofalcon` already ships these.

## Implementation notes / design decisions (feedback welcome)

- **`connector_id` and `connector_type` are `RequiresReplace`** — they're absent from the update request model (`DataconnectionmanagementUpdateDataConnectionRequest`), so they can't be changed in place.
- **Free-form connector config:** the API models `config.auth` and `config.params` as `interface{}` (their shape is connector-type-specific). Since the provider has no `jsontypes` precedent, I modeled them as **JSON-encoded strings** (validated in `ValidateConfig` and on apply), decoded into the generic API form. Open to a richer/typed schema if maintainers prefer.
- **Read merges only API-returned fields** — the read model doesn't echo input-only fields (`connector_id`, `log_sources`, `config`, etc.), so `wrap()` only sets the fields the API returns and preserves the rest from state.
- Standard `tferrors` handling, `utils.MarkdownDescription`, scopes block, and import-by-id, following CONTRIBUTING patterns. Scaffolded via `go run ./tools/generate resource`.

## Testing status

- ✅ `go build ./...`, `go vet ./...`, `gofmt`, and `go generate ./...` (docs) all pass locally.
- ⚠️ **Acceptance tests have NOT been run** — I don't have Falcon API credentials. The generated `*_resource_test.go` is a scaffold stub and needs fleshing out + a maintainer run of `make acctest` (`PKG=ngsiem_data_connection`) against a tenant. Opening as **draft** for that reason.
- The `config.auth`/`config.params` schema and the exact required API scope name are the most likely areas to need adjustment after a live run.

Happy to iterate on schema shape, naming, or split Create/Update config handling per maintainer guidance.

PR #412, opened 2026-06-20
feat(ngsiem_data_connection): add resource & data source

abdulharb:
## Summary

Adds the `crowdstrike_ngsiem_data_connection` resource and the `crowdstrike_ngsiem_data_connectors` data source for managing Next-Gen SIEM push-based data connections (e.g. HEC, Cribl) via an ingest URL + ingest token.

## What's included

- **Resource**: create/read/update/delete + import (composite `connector_id:connection_id`). `name`/`parser`/`description`/`log_sources` and the enrichment flags update in place; changing `connector_id` replaces the connection and rotates the token.
- **Data source**: lists connectors and resolves a connector ID by exact name.
- **Sweeper** for leaked acceptance-test connections.
- Docs + examples for both.

## Design decisions / gofalcon workarounds

- Ingest token is write-once (returned only at create); kept in state and `ImportStateVerifyIgnore`'d. Read warns when state has no token.
- Custom request bodies (`op.Params` hook) so an explicit `enable_*_enrichment = false` and a cleared `description` survive `omitempty`, and unset write-only fields are omitted so the merge-update server keeps prior values.
- Custom token reader (`op.Reader` hook): the token endpoint returns `resources` as a bare object that the generated array model can't decode; poll loop classifies 202/429/5xx as retry and malformed/200-with-errors as fail-fast.
- Push-only: pull connectors need config and are out of scope.

## Testing

Unit tests for pagination/dedup, body serialization, token decode + poll loop, wrap guards, and the sweep prefix guard. Acceptance tests for the full lifecycle (create/import, in-place updates, replace+rotate, `log_sources`), validation (plan-time `ExpectError`), and the invalid-parser 400 path. Verified green against a live Falcon us-2 tenant and compiles/tests clean against the current `main` (gofalcon v0.21.0); adds no new dependencies.


ffalor:
Hey @abdulharb thank you for trying to contribute I really appreciate it. Similar to what I said in #410 me not accepting this PR is not a reflection on the quality of the PR. It was oversight on my part.

Please let me know if you have any issues with the resource/data source. Thanks again for using the provider!

abdulharb:
Hey hey @ffalor!

All good! This PR was opened by Claude by accident while I was doing some testing in a private repo... it misunderstood me and pushed here instead. SORRY about that!

I also didn't love that I (Claude lol) needed to do some custom shimmies to get this working, so I've actually been bugging support about the gofalcon/OpenAPI mismatch on the token endpoint (case 02521275 if you're curious 🙂). Funny enough, the fix is coming out backwards from what I asked for lol. They're keeping the spec as-is and changing the API to actually return `resources` as an array, supposedly landing sometime next week. Which I THINK means your `regenerateTokenReader` override starts blowing up on every 200 the day that ships (it expects the object shape, same as mine did)... might be worth having it handle both shapes before then! Happy to share more from the case if it helps BUT I expect you can get more info than I can haha.

Anyway... at the end of the day, love that this got done and I'll definitely be testing it so I can hopefully push it into production soon! Love all the great work and thank you for improving the NG-SIEM connectors! (I'll probably start annoying you with future requests, or just opening my own PRs if I don't hit bugs that take me a month to fix hahaha)

ffalor:
@abdulharb thank's for the heads up! You're right If they change the shape then the current versions will start failing. I'll look into getting a new version out quickly so hopefully users of this resource can just upgrade. 

Looking forward to your future requests and feel free to open a PR whenever you want.

PR #426, opened 2026-07-13
Fix precedence policy

adelfazel:
The issue. When running terraform precedence policy in strict mode, there is a bug that causes terraform to incorrectly detect a nonexistent drift in state in applying.
````
error: Provider produced inconsistent result after apply
│ 
│ When applying changes to `crowdstrike_prevention_policy_precedence.prevention_policy_precedence, provider "provider[\"registry.terraform.io/crowdstrike/crowdstrike\"]" produced an unexpected new value: .ids: new element X has appeared.

````
Sample code:

```
platform_name: Linux
enforcement: strict
names:
  - 1
  - 2 
  - 3
```

What we believe is the issue:

When running on strict mode, getPreventionPoliciesByPrecedence reads CrowdStrike's full list hence incorrectly recognising a state drift due to existence of default policies.  
.

We suggest an alternative to this PR which when running on strict mode, QueryCombinedPreventionPoliciesParams() applies a different filter but above appears sufficient.
I reran the terraform plan with multiple prevention policies containing both default and non-default ones and the result was:
1. The above bug is not reprodueced.
2. Rerunning terraform doesn't create inconsistency

### DIFF
--- a/internal/prevention_policy/prevention_policy_precedence.go
+++ b/internal/prevention_policy/prevention_policy_precedence.go
@@ -190,10 +190,8 @@ func (r *preventionPolicyPrecedenceResource) Create(
 		return
 	}
 
-	if strings.EqualFold(plan.Enforcement.ValueString(), dynamicEnforcement) {
-		if len(policies) > len(plan.IDs.Elements()) {
-			policies = policies[:len(plan.IDs.Elements())]
-		}
+	if len(policies) > len(plan.IDs.Elements()) {
+		policies = policies[:len(plan.IDs.Elements())]
 	}
 
 	plan.LastUpdated = utils.GenerateUpdateTimestamp()
@@ -218,10 +216,8 @@ func (r *preventionPolicyPrecedenceResource) Read(
 		return
 	}
 
-	if strings.EqualFold(state.Enforcement.ValueString(), dynamicEnforcement) {
-		if len(policies) > len(state.IDs.Elements()) {
-			policies = policies[:len(state.IDs.Elements())]
-		}
+	if len(policies) > len(state.IDs.Elements()) {
+		policies = policies[:len(state.IDs.Elements())]
 	}
 
 	resp.Diagnostics.Append(state.wrap(ctx, policies)...)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ffalor** [comment]
> When running on strict mode, getPreventionPoliciesByPrecedence reads CrowdStrike's full list hence incorrectly recognising a state drift due to existence of default policies.

The default policy is removed from the list we get back on read, we sort by precedence and remove the last element from the returned call. This should be removing the default policy from the list. When you define the resource in your config you should not include the default policy ID.

https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/eb1127df551f2a0afe5970f162e5ded9566b9bda/internal/prevention_policy/prevention_policy_precedence.go#L331-L337

However, it looks like the read call does not paginate, the default limit is 100 so if you have more than 100 prevention policies you could run into a bug. 

~Do you have over 100 prevention policies? Could that be the cause of what you are seeing?~

This error, produced an unexpected new value: .ids: new element X has appeared, happens when the resource sets the precedence and then reads the policies back into state, and the read-back contains more ids than were provided in ids.

Under strict enforcement the API requires the complete set of non-default policies for the platform. If you provide fewer, SetPreventionPoliciesPrecedence rejects the request with a 400 before we ever read anything back, so you can't hit this error just by under-specifying ids.

The scenario I can see producing it is a race: if a new prevention policy is created for that platform after we set precedence but before we read the list back, the read-back returns one more id than was in your plan, and Terraform flags it as a new element appearing in state.

Is this error consistent, or does it only happen intermittently? And was any prevention policy created (by another process, another apply, or in the console) around the same time as this run?

**Comment by adelfazel** [comment]
We have ~10 policies ~5 of which were non-default and were all listed in the request. 
The issue is consistent, we tested in two different cids and both cases we can replicate the issue. Above code actually fixes it.  

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module._policy_precedence, provider "provider[\"registry.terraform.io/crowdstrike/crowdstrike\"]" produced an unexpected new value: .ids: new element 9 has appeared.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵

ffalor:
Hey @adelfazel are you in a flight control aware cid? I just merged #435 and released v0.0.82 yesterday. It now correctly strips all defaults using the name, and it also filters the response to only the CID that is being managed (requires a new scope). This seems to be the behavior you are seeing, can you see if v0.0.81 resolves your issue?

The prevention policies API returns every policy visible to the account, including ones that live in parent/child CIDs that can't be reordered and don't appear in the console. To handle this correctly, the provider now looks up the caller's own CID and filters the response down to just that CID's policies, ignoring the rest. Resolving that CID is what requires the additional Sensor Download: Read scope.

PR #450, opened 2026-07-21
fix(firewall): deterministic rule order on import via rule group rule_ids

### LINKED ISSUE #449: firewall_rule_group: nondeterministic rule order on import prevents clean import plans
## Summary

`crowdstrike_firewall_rule_group` stores rules in a nondeterministic order when the resource is imported, which makes it impossible to get a clean (`0 to change`) plan when onboarding existing rule groups with `import` blocks or `terraform import`. Every plan after an import shows spurious rule reordering diffs.

## Root cause

`GET /fwmgr/entities/rules/v1` returns rules in a nondeterministic order. Three identical consecutive requests for the same 11 rule IDs returned three different orderings (us-2 cloud):

```
run 1: ['Allow Prometheus metrics', 'Falcon Sensor Cloud Communication', 'Allow ICMPv4', ...]
run 2: ['Allow ICMPv4', 'Allow Multicast IPv6', 'Allow Localhost Communications', ...]
run 3: (different again)
```

The provider already compensates for this during normal reads: `orderRulesByPlanNames()` in `internal/firewall/rule_group_helpers.go` re-sorts the fetched rules to match the plan order by rule name. But when there is no plan to sort against, i.e. `terraform import` / `import {}` blocks, it falls through to "return as-is" and the random API response order is persisted to state:

```go
// If no plan rules provided, return as-is
if !utils.IsKnown(planRules) {
    return apiRules, diags
}
```

(`ImportState` is `ImportStatePassthroughID`, so the subsequent read runs with no prior rules in state.)

## Impact

- Importing any rule group with more than one rule produces a plan diff that shuffles rules (order is semantic, it's the rule precedence), so import-onboarding workflows that gate on `0 to change` cannot pass.
- Since the read path orders by prior state once imported, the random imported order is sticky; converging requires applying a reorder update (a real PUT to the rule group).

Single-rule groups import cleanly, which confirms the ordering is the only issue.

## Repro

1. In the Falcon console, create a firewall rule group with 3+ rules.
2. Write matching Terraform config with the rules in the console's precedence order.
3. `import { to = crowdstrike_firewall_rule_group.example, id = "<group id>" }`
4. `terraform plan` shows the rule list being reordered (which rules move varies run to run).

## Proposed fix

The rule group object itself carries the canonical precedence in `rule_ids` (entries reference the rules' `family` identifiers). Ordering fetched rules by the group's `rule_ids` before the plan-name ordering makes reads deterministic in all cases, including import:

```go
canonicalRules := orderRulesByRuleIDs(rulesResult.Payload.Resources, ruleGroup.RuleIds)
orderedRules, d := orderRulesByPlanNames(ctx, canonicalRules, planRules)
```

PR with this fix (+ unit tests) incoming. Verified against a live us-2 tenant: importing 12 existing rule groups (188 rules total) went from 10-of-12 showing reorder diffs to all 12 importing with zero changes.

Provider version: 0.0.80 (bug present on current `main`)
Terraform: 1.9.8

lennyvaknine43:
## Summary

Fixes #449.

`GET /fwmgr/entities/rules/v1` returns rules in a nondeterministic order (verified across identical consecutive requests on us-2). Normal reads already compensate via `orderRulesByPlanNames()`, but reads with no plan available — `terraform import` / `import {}` blocks — persisted the random response order to state. Importing any rule group with more than one rule produced spurious rule-reordering diffs, so import onboarding gated on `0 to change` could never pass.

## Change

- Add `orderRulesByRuleIDs()`: canonicalizes fetched rules to the rule group's `rule_ids` order, which is the group's precedence order. `rule_ids` entries reference the rules' `family` identifiers (stable across rule updates), so matching is by `Family` with a fallback to `ID`; unreferenced rules (shouldn't happen) are appended in response order.
- `readRuleGroupState()` canonicalizes first, then applies the existing plan-name ordering, preserving current behavior for normal reads while making no-plan reads (import) deterministic.
- Unit tests for the new ordering function (family matching, ID fallback, unreferenced-rule append, nil/duplicate handling, empty inputs).

## Verification

- `go build ./...`, `go vet ./internal/firewall/`, new unit tests pass.
- Live tenant (us-2): importing 12 existing rule groups (188 rules total) with `import {}` blocks went from 10-of-12 groups showing reorder diffs to **all 12 importing with zero changes**, with config rules written in `rule_ids` order. Single-rule groups (unaffected by ordering) were clean before and after, confirming ordering was the only diff source.


ffalor (approved):
Thanks for finding and creating a fix for this!

redhatrises (approved):

PR #455, opened 2026-07-23
fix(response_policy_precedence): fix case insensitivity

robertzakrocki:
Currently the `platform_name` in crowdstrike_response_policy_precedence is case sensitive causing the terraform plan to fail if one uses the example code provided. eg (https://registry.terraform.io/providers/CrowdStrike/crowdstrike/latest/docs/resources/response_policy_precedence).

```
...
platform_name = "linux"
...
```

Terraform throws this error if lower case is used:

```
╷
│ Error: Invalid Attribute Value Match
│ 
│   with crowdstrike_response_policy_precedence.response_policy_precedence,
│   on response_policies.tf line 42, in resource "crowdstrike_response_policy_precedence" "response_policy_precedence":
│   42:   platform_name = "linux"
│ 
│ Attribute platform_name value must be one of: ["Windows" "Linux" "Mac"], got: "linux"
```

### DIFF
--- a/internal/response_policy/response_policy_precedence.go
+++ b/internal/response_policy/response_policy_precedence.go
@@ -141,7 +141,7 @@ func (r *responsePolicyPrecedenceResource) Schema(
 				Required:    true,
 				Description: "The platform of the response policies. One of: Windows, Mac, Linux",
 				Validators: []validator.String{
-					stringvalidator.OneOf("Windows", "Linux", "Mac"),
+					stringvalidator.OneOfCaseInsensitive("Windows", "Linux", "Mac"),
 				},
 			},
 			"last_updated": schema.StringAttribute{


ffalor:
Thanks for catching this! This is really a documentation inconsistency, so I'd like to fix it in the example rather than in the validator.

The API expects the platform name in Title case (`Windows`, `Mac`, `Linux`). We used to normalize casing for the user, but we've moved away from that in favor of validating the exact value up front, so keeping the strict `OneOf` here is intentional. The `crowdstrike_response_policy` resource itself uses the strict validator too, so loosening the precedence resource would put the two out of sync.

Instead, the fix would be to correct the casing in the example. In `examples/resources/crowdstrike_response_policy_precedence/resource.tf`:

```diff
-  platform_name = "linux"
+  platform_name = "Linux"
```

Then run `make gen` to regenerate the docs (the registry page linked in the issue is generated from that example file).

Since you found the original issue, I wanted to check whether you'd like to update the PR to take this approach and keep the credit. Happy to make the change myself if you'd prefer, just let me know!

robertzakrocki:
Many thanks for clarifying. I initially assumed this was just a documentation issue, but after comparing `response_policy_precedence.go` with the other policy precedence files, I noticed that they normalise the user casing - so that’s why I opened the PR.

For example:

https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/3cd421f60d5e8940b2ddc0c38acb1ce6fd4ecbbc/internal/prevention_policy/prevention_policy_precedence.go#L154

https://github.com/CrowdStrike/terraform-provider-crowdstrike/blob/3cd421f60d5e8940b2ddc0c38acb1ce6fd4ecbbc/internal/sensor_update_policy/sensor_update_policy_precedence.go#L150

However, I didn’t realise you were in the process of moving away from that approach.

As per my PR, it’s fine either way - please go ahead and make the changes you consider necessary, if you’re okay with that.

ffalor:
That makes complete sense. I hadn't realized all the precedence resources so far use `OneOfCaseInsensitive`; I was looking at the actual policy resources, which all use `OneOf`, so that's what I was aligning with.

We'll go with your change. I need to give some more thought to how we want to standardize this across the policy and precedence resources. Thanks again for catching it and for the fix!

ffalor (approved):
LGTM. Thanks again for the fix!

PR #463, opened 2026-07-28
fix: clear default request timeout in ml exclusion update params

### LINKED ISSUE #462: ml_file_path_exclusion: updates fail with context deadline exceeded at 30s for large host_groups lists
🤖 *posted by claude in collaboration with claire*

## Summary

`crowdstrike_ml_file_path_exclusion` updates fail with:

```
Error: Failed to update

Patch "https://api.crowdstrike.com/policy/entities/ml-exclusions/v1": context
deadline exceeded
```

whenever the PATCH takes longer than 30 seconds — which reliably happens for exclusions with large `host_groups` lists (~36 groups in our tenant). The failure is deterministic, not transient: the same resources fail at exactly 30s on every retry. The cap is not configurable from Terraform config and does not come from the plugin framework; it is go-openapi's `client.DefaultTimeout`, which this resource's update path fails to clear.

## Root cause

The update path replaces the generated SDK params with a custom writer, to work around gofalcon's `UpdateMLExclusionsV1Params` using the wrong body model (`internal/ml_file_path_exclusion/ml_file_path_exclusion.go`):

```go
params := ml_exclusions.NewUpdateMLExclusionsV1ParamsWithContext(ctx)
updateResp, err := r.client.MlExclusions.UpdateMLExclusionsV1(
	params,
	func(operation *runtime.ClientOperation) {
		// SDK update params use the wrong body model (SvExclusionsUpdateReqV1),
		// so we override the request writer with the ML-compatible payload.
		operation.Params = &mlFilePathExclusionUpdateParams{Body: updateReq}
	},
)
```

Generated go-swagger params begin `WriteToRequest` with `r.SetTimeout(o.timeout)`; with the `WithContext` constructor `o.timeout` is zero, so the per-request timeout is cleared and the call is bounded only by the operation context. The custom writer only sets the body:

```go
func (p *mlFilePathExclusionUpdateParams) WriteToRequest(r runtime.ClientRequest, _ strfmt.Registry) error {
	if p.Body != nil {
		if err := r.SetBodyParam(p.Body); err != nil {
			return err
		}
	}

	return nil
}
```

go-openapi initializes every request with `timeout: DefaultTimeout` (30s, `go-openapi/runtime/client/request.go`), and since nothing overrides it, `Submit` wraps the operation context in `context.WithTimeout(parentCtx, 30*time.Second)`. Result: a hard 30-second deadline on exactly this one call.

The PATCH latency on this endpoint scales with the size of `groups`. In the same applies: a 12-group update completed in 22s, while four ~36-group updates all exceeded 30s, every attempt. Other resources confirm the cap is specific to the overridden params: `crowdstrike_sensor_visibility_exclusion` updates in the same run took 39s+ without error (generated params clear the timeout), and `crowdstrike_ioa_exclusion` updates carrying 35+ groups completed in 4–9s (different, faster endpoint).

## Impact

- Any `ml_file_path_exclusion` whose update PATCH takes >30s can never be updated through Terraform; retries fail identically.
- The API often finishes processing the PATCH after the client disconnects, so the change lands server-side while Terraform records a failure; the next refresh reconciles it. This makes the bug look like transient flakiness when it's a deterministic client-side deadline.
- `crowdstrike_ml_certificate_exclusion` uses the same pattern (`op.Params = &certExclusionsUpdateParams{...}` in `internal/ml_certificate_exclusion/ml_certificate_exclusion.go`, also without a timeout reset), so its updates carry the same cap.
- Create/Delete paths use the generated params and are unaffected.

## Repro

1. Create a `crowdstrike_ml_file_path_exclusion` with a few host groups.
2. Update its `host_groups` to a list large enough that the PATCH exceeds 30s (~36 groups in our tenant, us-1; the threshold presumably varies by tenant/cloud).
3. `terraform apply` fails with `context deadline exceeded` at exactly 30s elapsed, on every attempt.

## Proposed fix

Have the custom writers mirror the generated params' timeout handling — clear the default at the top of `WriteToRequest` so the call is bounded by the operation context, exactly as the generated params behave:

```go
if err := r.SetTimeout(0); err != nil {
	return err
}
```

Same change for `certExclusionsUpdateParams` in `internal/ml_certificate_exclusion`. PR with this fix (+ unit tests) incoming.

Provider version: 0.0.79 (code unchanged on current `main` / v0.0.83)
Terraform: 1.13.4

yoclaire:
## Summary

Fixes #462.

`crowdstrike_ml_file_path_exclusion` and `crowdstrike_ml_certificate_exclusion` updates replace the generated SDK params with custom writers (to work around the SDK body models). The generated `WriteToRequest` implementations start with `r.SetTimeout(o.timeout)` — zero for params built with the `WithContext` constructor — which clears go-openapi's 30s `DefaultTimeout`. The custom writers skip that call, so every update PATCH for these two resources carries a hard, non-configurable 30-second deadline and fails with `context deadline exceeded` whenever the API takes longer. This is reliably reproducible with large `host_groups` lists; see #462 for the full analysis.

This change clears the per-request timeout in both custom writers so the request is bounded by the operation context, exactly like the generated params the writers replaced. `SetTimeout(0)` is the no-deadline path in go-openapi's `Submit`: `request.timeout == 0` wraps the operation context with `context.WithCancel` instead of `context.WithTimeout`.

## Changes

- `internal/ml_file_path_exclusion`: call `SetTimeout(0)` in `mlFilePathExclusionUpdateParams.WriteToRequest`
- `internal/ml_certificate_exclusion`: call `SetTimeout(0)` in `certExclusionsUpdateParams.WriteToRequest`
- unit tests asserting both writers clear the request timeout and still set the body

## Testing

- `go build ./...` and `go test ./internal/ml_file_path_exclusion/... ./internal/ml_certificate_exclusion/...` pass.
- The failure mode was observed on a live us-1 tenant running provider v0.0.79: `ml_file_path_exclusion` updates with ~36-group `host_groups` lists fail at exactly 30s on every retry, while `sensor_visibility_exclusion` updates in the same apply ran 39s+ without error (generated params, no cap).

### DIFF
--- a/internal/ml_certificate_exclusion/ml_certificate_exclusion.go
+++ b/internal/ml_certificate_exclusion/ml_certificate_exclusion.go
@@ -655,6 +655,12 @@ type certExclusionsUpdateParams struct {
 }
 
 func (p *certExclusionsUpdateParams) WriteToRequest(r runtime.ClientRequest, _ strfmt.Registry) error {
+	// Generated params clear the per-request timeout via SetTimeout; without it
+	// the request is capped at go-openapi's 30s DefaultTimeout, which updates
+	// with large host_groups lists can exceed.
+	if err := r.SetTimeout(0); err != nil {
+		return err
+	}
 	if p.Body == nil {
 		return nil
 	}
--- a/internal/ml_certificate_exclusion/ml_certificate_exclusion_test.go
+++ b/internal/ml_certificate_exclusion/ml_certificate_exclusion_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/crowdstrike/gofalcon/falcon/models"
 	"github.com/crowdstrike/terraform-provider-crowdstrike/internal/utils"
+	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/strfmt"
 	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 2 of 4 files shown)


ffalor (approved):
I appreciate that fix! It looks like the same issue exists in a few other places.

This will be included in tomorrows release.

PR #479, opened 2026-08-11
feat(data_protection): add data_protection_policy resource

ffalor:
Manages a Falcon Data Protection policy for a single platform: its settings, assigned host groups, and assigned classifications. platform_name is immutable and forces replace. The resource is the authoritative definition of the policy, so every setting is written on each apply and any setting whose remote value differs from the configuration is reverted.

tferrors now extracts error detail from typed gofalcon response payloads in addition to runtime.APIError bodies, so an API rejection surfaces its real message instead of a bare status. Adds IsServerError, a runtime.ClientResponseStatus check, so callers can tell a retryable 5xx from a request the API refused.

framework/validators gains BoolRequiresBool, AttributeRequiresBool and BoolRequiresStringValue for the policy's requires-another-setting dependencies, plus SetJoinedLengthAtMost for the 2048 character limit on the comma-joined be_exclude_domains value.

Fixes https://github.com/CrowdStrike/terraform-provider-crowdstrike/pull/306


redhatrises (approved):

ffalor:
Fixes #284

PR #491, opened 2026-08-26
feat(cloud_azure_tenant): add crowdstrike_cloud_azure_tenant data source

ffalor:
Adds a data source that looks up a single Azure tenant registered in Falcon Cloud Security by either its Azure tenant ID or its Falcon registration ID, exposing the same attributes as the managed resource plus the tenant's Event Hub settings.

Computed collections return empty rather than null so callers can use length() and for_each on the result without guarding with coalesce, and they are lists rather than sets since they are never diffed against config and indexing the result is useful.

The product feature and Event Hub settings flattening previously lived inline in the resources' wrap methods; both are now shared helpers used by the resources and the data source. Flattening no longer dereferences API pointers that the spec allows to be nil.


ffalor:
CI failure is unrelated to the change and related to the go tooling change, a separate PR will be created to fix instead of burying the changes in this PR with unrelated lint fixes.

This PR is ready for review.

redhatrises (approved):

PR #495, opened 2026-08-28
fix(firewall_rule_group): edit rules in place and stop perpetual rule diffs

ffalor:
Rule updates now patch the existing rule instead of removing and re-adding it, so a rule keeps the Rule ID shown in the console and in firewall events across edits, renames, and moves. Rules are matched by content, so reordering or inserting a rule no longer recreates the unchanged rules around it. An out-of-band reorder is detected and reported instead of patching the wrong rules.

Several updates that silently did nothing now apply: clearing executable_path or service_name, turning off watch_mode, switching a rule off ICMP, and clearing fqdn. Address and port list edits are rebuilt element by element, since a whole-list replace appended on this API and an intermediate empty address list could widen a rule to "any address".

"Any" now has one spelling. Omitted address lists default to a single {address = "*", netmask = 0} entry and icmp_type/icmp_code default to "*" on ICMP rules and null elsewhere, matching what the API reports, so configuring "*" and omitting the attribute produce identical state instead of diffing forever. Single-port ranges keep their configured end value, an explicitly empty rules list is preserved, and a rule's id in state is the stable family id that matches the group's precedence order.

Rule groups with more than 100 rules are readable again; rule IDs are fetched in batches. Soft-deleted rule groups are treated as not found so Terraform plans a recreate, a server-side-disabled fqdn reads back as unset, and unresolvable rule IDs or malformed address and port entries raise explicit errors rather than a skewed rules list.

New plan-time validation rejects configurations the API cannot honor: ICMP attributes on non-ICMP rules, an address_family that contradicts an ICMP protocol, specific addresses when address_family is ANY, a non-zero netmask on the "*" address, a port range whose end is not greater than its start, empty strings and empty lists where the attribute must be omitted, and a non-ANY network_location on Linux.


ffalor:
Ignore CI failures, they are fixed in main, the workflows in this branch havent been updated.

Pushing a fix for the merge conflict now. This PR is ready for review.

ffalor:
Related to #474

carlosmmatos (approved):

carlosmmatos (approved):

redhatrises (approved):

Issue #493, opened 2026-08-28
Custom IOM rule inheriting a SubDomain "IOM" parent fails on PATCH ("Cloud is not a supported platform for SubDomain IOM") when assigning custom controls — no HCL workaround possible


Klorsche:
## Summary

When creating a `crowdstrike_cloud_security_iom_custom_rule` that (a) inherits from a parent rule via `parent_rule_id` and (b) assigns a custom compliance control via `controls`, the provider performs a `POST` to create the rule followed by a `PATCH` to apply the controls.

- The **`POST` succeeds** and the rule is created on Falcon **with the custom control correctly assigned**.
- The subsequent **`PATCH` fails** with `400 Bad Request`: Cloud is not a supported platform for SubDomain IOM

Because the `PATCH` fails, Terraform marks the apply as failed and **does not persist the resource into state**, even though the resource exists and is fully correct on Falcon. This makes the resource unmanageable via a normal `terraform apply` and breaks any CI/CD automation. We have verified that **no `lifecycle.ignore_changes` configuration can work around this**, because the failing `PATCH` is part of the *create* sequence, not a subsequent update.

## Terraform and Provider Versions

- Terraform: `v1.15.8` (on `windows_amd64`)
- Provider `crowdstrike/crowdstrike`: reproduced on **`0.0.83`** and **`0.0.86`**
- Falcon Cloud Region: `eu-1`

## Affected Resource

- `crowdstrike_cloud_security_iom_custom_rule`

## Terraform Configuration

```hcl
resource "crowdstrike_cloud_security_iom_custom_rule" "this" {
name           = "custom_aws_s3_no_public_access"
description    = "Verify that the S3 bucket does not allow public access."
cloud_provider = "AWS"
resource_type  = "AWS::S3::Bucket"
parent_rule_id = "REDACTED"  # default CSPM IOM S3 rule

controls = [
  {
    authority = "Custom"
    code      = "REDACTED"  # custom control code from a custom compliance framework
  }
]
}
```

## Parent rule details

The parent rule is a default CSPM IOM rule. Querying it via the API returns (relevant fields):

- domain: CSPM
- subdomain: IOM
- provider: AWS
- rule_logic_list[].platform: Cloud

The error Cloud is not a supported platform for SubDomain IOM appears to relate to the combination of subdomain: IOM and platform: Cloud being rejected by the update (PATCH) endpoint, even though the same rule is accepted by the create (POST) endpoint.

[...]

PR #496, opened 2026-08-29
fix(cloud_security): keep custom rule ids and fix inherited rule updates

ffalor:
Creating a custom IOM rule from a parent rule could fail while leaving the rule behind in Falcon with nothing recorded in Terraform state.

The follow-up update issued during create sent back the platform value the API reports for an inherited rule, "Cloud", which the update endpoint rejects with "Cloud is not a supported platform for SubDomain IOM". Create and update now both send cloud_provider, the only value the API accepts, and cloud_platform is read from the API without ever being sent back.

When that update failed, the newly created rule was discarded along with its id, so the apply errored with no resource in state and the rule could only be recovered by importing it by hand. The id is now recorded as soon as the API reports one, before any follow-up call, and a create that returns a rule alongside body-level errors keeps its id as well.

Refs #493

### DIFF
--- a/internal/cloud_security/custom_iac_rule_resource.go
+++ b/internal/cloud_security/custom_iac_rule_resource.go
@@ -256,6 +256,10 @@ func (r *cloudSecurityIacCustomRuleResource) Create(
 
 	rule, diags := r.createCloudPolicyRule(ctx, &plan)
 	resp.Diagnostics.Append(diags...)
+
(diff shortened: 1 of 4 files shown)


ffalor:
Superseded by #500, which includes this commit (e910a93) unchanged and builds on it with custom controls support. Closing in favor of #500.

PR #501, opened 2026-09-15
feat(cloudregistration): use existing wif pool for GCP

mzastavnyi-cs:
- Adds a new optional existing_wif_pool_id attribute to crowdstrike_cloud_google_registration, letting a project-scoped registration attach to a Workload Identity Federation pool owned by another registration under the same CID.
- wif_project / wif_project_number become Optional+Computed — still required unless existing_wif_pool_id is set, in which case they're resolved from the owner registration.
- Adds ValidateConfig checks rejecting existing_wif_pool_id when combined with organization/folders scope, realtime_visibility, dspm, or vulnerability_scanning enabled.
- Adds a new computed wif_pool_registration_id attribute exposing which registration owns the attached pool.


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/crowdstrike/terraform-provider-crowdstrike
 go 1.24.3
 
 require (
-	github.com/crowdstrike/gofalcon v0.21.1-0.20260714161657-21b34ca8aee1
+	github.com/crowdstrike/gofalcon v0.21.1-0.20260911170427-cef8bd0ef61e
```
mzastavnyi-cs: PR request is blocked until gofalcon PR is merged into main.

alex-grant-cs (approved):

internal/cloud_google_registration/cloud_google_registration_settings.go
```diff
@@ -820,8 +843,16 @@ func (r *cloudGoogleRegistrationSettingsResource) updateRegistration(
...
+	}
+
+	// FrameworkToStringPointer turns null/unknown into a pointer to "", which the backend treats
+	// as "WIF properties provided" (even though empty) and rejects for registrations attached to
+	// another registration's WIF pool.
+	if !data.WifPoolName.IsNull() && !data.WifPoolName.IsUnknown() {
+		updateReq.WifPoolName = flex.FrameworkToStringPointer(data.WifPoolName)
+	}
+	if !data.WifProviderName.IsNull() && !data.WifProviderName.IsUnknown() {
+		updateReq.WifProviderName = flex.FrameworkToStringPointer(data.WifProviderName)
```
ffalor: How is someone supposed to "unset" one of these values if patch sends nothing if they omit the attribute?

internal/cloud_google_registration/cloud_google_registration.go
```diff
@@ -499,6 +510,13 @@ func (r *cloudGoogleRegistrationResource) Schema(
 					),
 				},
 			},
+			"existing_wif_pool_id": schema.StringAttribute{
+				Optional:    true,
```
ffalor, line 520: `utils.IsKnown` is false for unknown, so the switch in `ModifyPlan` treats an unresolved `existing_wif_pool_id` as absent. With the usage the example documents (`existing_wif_pool_id = crowdstrike_cloud_google_registration.owner.wif_pool_id`), adding the attach to an existing registration plans as an in-place update rather than a replacement.

`stringplanmodifier.RequiresReplace()` already skips create and destroy and compares `PlanValue` to `StateValue` directly, so null to known, known to null, known to known, and unknown against a null state all behave correctly:

```suggestion
				Optional:    true,
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.RequiresReplace(),
				},
```

internal/cloud_google_registration/cloud_google_registration.go
```diff
@@ -1054,6 +1139,30 @@ func (r *cloudGoogleRegistrationResource) ModifyPlan(ctx context.Context, req re
...
+	}
+
+	if existingWifPoolIDChanged {
+		resp.RequiresReplace = append(
+			resp.RequiresReplace,
+			path.Root("existing_wif_pool_id"),
+			path.Root("wif_project"),
+			path.Root("wif_project_number"),
+		)
+	}
```
ffalor: With `stringplanmodifier.RequiresReplace()` on the attribute (comment above), this whole block can go. The built-in modifier covers every transition this switch computes, and handles the unknown case correctly.

Dropping the `wif_project` / `wif_project_number` paths only changes which attributes the plan output annotates with `# forces replacement`; any non-empty `RequiresReplace` replaces the whole resource, and citing only the attribute that actually changed is arguably more accurate.

```suggestion
```

internal/cloud_google_registration/cloud_google_registration.go
```diff
@@ -499,6 +510,13 @@ func (r *cloudGoogleRegistrationResource) Schema(
 					),
 				},
 			},
+			"existing_wif_pool_id": schema.StringAttribute{
+				Optional:    true,
+				Description: "The ID of an existing GCP Workload Identity Pool, owned by another registration under the same CID, to attach this registration to instead of creating a new pool. Only valid for project-scoped registrations with no real-time visibility, DSPM, or vulnerability scanning enabled",
```
ffalor, line 522: TLDR: can existing_wif_pool_id be set in wrap? Does the api return it?

`ImportState` is a bare `ImportStatePassthroughID` and `wrap` never assigns `m.ExistingWifPoolID`, so after importing an attached registration this is null in state. The next plan then sees null to known and `terraform apply` destroys and recreates a live cloud registration.

`wrap` should populate it when `WifPoolRegistrationID != ""`, presumably from `WifProperties.PoolID`. Worth checking whether the API returns the pool ID in the same format it accepts on create; if it normalizes the value, the attribute also needs to be `Computed` to avoid "Provider produced inconsistent result after apply".

internal/cloud_google_registration/cloud_google_registration.go
```diff
@@ -701,6 +785,7 @@ func (r *cloudGoogleRegistrationResource) Create(
 
 	createReq.ResourceNameSuffix = flex.FrameworkToStringPointer(plan.ResourceNameSuffix)
 	createReq.ResourceNamePrefix = flex.FrameworkToStringPointer(plan.ResourceNamePrefix)
+	createReq.ExistingWifPoolID = flex.FrameworkToStringPointer(plan.ExistingWifPoolID)
```
ffalor: Question: `flex.FrameworkToStringPointer` returns a pointer to `""` for null and unknown (`internal/framework/flex/string.go:31-40`), never nil, and `omitempty` on a `*string` only skips a nil pointer. So `"existing_wif_pool_id": ""` is now serialized on every GCP registration create, including for everyone who isn't using this feature.

Is an empty string fine for the backend here, or should we omit the field when it isn't set? Asking because the comment you added at `cloud_google_registration_settings.go:848-850` says this backend reads an empty-but-present WIF field as "provided".

internal/cloud_google_registration/cloud_google_registration.go
```diff
@@ -477,8 +480,12 @@ func (r *cloudGoogleRegistrationResource) Schema(
 				},
 			},
 			"wif_project": schema.StringAttribute{
-				Required:    true,
-				Description: "The Google Cloud project ID for Workload Identity Federation",
+				Optional:    true,
+				Computed:    true,
```
ffalor, line 490: Making this Optional+Computed means `Update` now sends the owner's WIF project for an attached registration. The chain: the backend mirrors the owner's WIF values into GET (the premise of the settings changes in this PR), `wrap` copies them into state, and with a null config plus `UseStateForUnknown` the planned value is that mirrored value. `Update` at line 965-967 then sends `WifProjectID`/`WifProjectNumber` unconditionally, and `DtoUpdateGCPRegistrationRequest` has no `existing_wif_pool_id`, so the backend has no way to know this registration doesn't own the pool.

Net effect: a plain rename of an attached registration submits a claim on the owner's WIF project. Same class of bug the settings resource had to fix for `wif_pool_name`/`wif_provider_name`, but for `wif_project`/`wif_project_number` it isn't handled. No test covers it, since the only attach test is skipped and has a single step.

internal/cloud_google_registration/cloud_google_registration.go
```diff
@@ -146,6 +146,8 @@ type cloudGoogleRegistrationResourceModel struct {
 	WifProviderID               types.String `tfsdk:"wif_provider_id"`
 	WifProviderName             types.String `tfsdk:"wif_provider_name"`
 	WifIdentitySource           types.String `tfsdk:"wif_identity_source"`
+	ExistingWifPoolID           types.String `tfsdk:"existing_wif_pool_id"`
+	WifPullRegistrationID       types.String `tfsdk:"wif_pool_registration_id"`
```
ffalor: Typo: `WifPullRegistrationID` should be `WifPoolRegistrationID`. Also at line 273.

internal/cloud_google_registration/cloud_google_registration_settings.go
```diff
@@ -625,11 +629,30 @@ func (r *cloudGoogleRegistrationSettingsResource) Delete(
 		return
 	}
 
+	current, diags := r.getRegistration(ctx, data.RegistrationID.ValueString())
+	resp.Diagnostics.Append(diags...)
+	if tferrors.HasNotFoundError(resp.Diagnostics) {
+		return
+	}
```
ffalor, line 653: This turns an already-deleted registration into a destroy failure. `diags` is appended to `resp.Diagnostics` before the `HasNotFoundError` check, and `NewNotFoundError` builds an error diagnostic (`internal/tferrors/errors.go:34-36`), so the early return doesn't swallow it: `terraform destroy` fails with "No registration found for registration ID" instead of succeeding. If the parent registration is deleted out of band, the settings resource can no longer be removed from state.

`Read` at line 547-559 has the right shape, testing the local `diags` before appending:

```suggestion
	current, diags := r.getRegistration(ctx, data.RegistrationID.ValueString())
	if tferrors.HasNotFoundError(diags) {
		return
	}
	resp.Diagnostics.Append(diags...)
```

internal/cloud_google_registration/cloud_google_registration_settings.go
```diff
@@ -207,15 +207,19 @@ func (r *cloudGoogleRegistrationSettingsResource) Schema(
 				},
 			},
 			"wif_pool_name": schema.StringAttribute{
-				Optional:    true,
-				Description: "The Workload Identity Federation (WIF) pool name.",
+				Optional: true,
+				Computed: true,
```
ffalor: Adding `Computed` here (and on `wif_provider_name` at line 219-222) creates two problems.

1. Nothing can be unset any more, for anyone. Terraform keeps the prior state value for an Optional+Computed attribute with a null config, so removing `wif_pool_name` from config now produces no diff at all. Previously (Optional only) it planned null, `FrameworkToStringPointer` sent `""`, and the backend cleared it.

2. The attached-registration guard only holds on the first write. A null config means unknown on create, so the new check at line 851 correctly omits the field, but `wrap` at line 376-377 then stores the backend's mirrored owner pool name in state. On the next update (changing `log_ingestion_sink_name`, say) the planned value is that known non-null string, the guard passes it through, and you send `wif_pool_name` for an attached registration: exactly the request the comment at line 641-643 says the backend rejects.

`Computed` is fighting a drift problem created elsewhere, namely `wrap` storing values this resource doesn't own. Keeping these Optional-only and leaving them null in `wrap` when `WifPoolRegistrationID != ""`, then skipping them in `updateRegistration` on the same condition, avoids both. `Delete` already does that check.

internal/cloud_google_registration/cloud_google_registration_test.go
```diff
@@ -845,3 +845,202 @@ resource "crowdstrike_cloud_google_registration" "test" {
...
+				Config: testAccCloudGoogleRegistrationConfig_project(rName, infraProjectID, wifProjectID, wifProjectNumber, projectID),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("name"), knownvalue.StringExact(rName)),
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("wif_project"), knownvalue.StringExact(wifProjectID)),
+				},
+			},
+			{
+				Config:             testAccCloudGoogleRegistrationConfig_existingWifPoolID(rName, projectID, infraProjectID),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
```
ffalor, line 926: `PlanOnly` + `ExpectNonEmptyPlan` only asserts that the plan isn't empty, which is equally true of an in-place update, so this test passes whether or not the replacement actually happens. It wouldn't catch the `ModifyPlan` issue in the other comment.

Note that `ConfigPlanChecks.PreApply` is skipped entirely for a `PlanOnly` step (`testing_new_config.go:108-149`); `PostApplyPreRefresh` is the one that runs (`testing_new_config.go:299-309`):

```suggestion
				PlanOnly:           true,
				ExpectNonEmptyPlan: true,
				ConfigPlanChecks: resource.ConfigPlanChecks{
					PostApplyPreRefresh: []plancheck.PlanCheck{
						plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionReplace),
					},
				},
```

internal/cloud_google_registration/cloud_google_registration_test.go
```diff
@@ -845,3 +845,202 @@ resource "crowdstrike_cloud_google_registration" "test" {
...
+	projectID := generateGoogleCloudProjectID()
+	infraProjectID := generateGoogleCloudProjectID()
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: acctest.ProtoV6ProviderFactories,
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccCloudGoogleRegistrationConfig_existingWifPoolID(rName, projectID, infraProjectID),
+				ExpectError: regexp.MustCompile(`(?s).*`),
```
ffalor: `(?s).*` matches any error at all, including a credential failure or an unrelated plan error, so this test can pass for entirely the wrong reason. Worth pinning to the actual expected backend message once the feature is available, or dropping the test until then.

ffalor (dismissed):
Reviewed the attach-to-existing-WIF-pool change. The feature shape makes sense, and `existing_wif_pool_id` has to be a create-time input rather than something derived, since it changes what the backend does rather than just reporting a value.

A few things to work through, in rough priority order:

- Create unconditionally serializes `existing_wif_pool_id: ""` for every registration, including non-attach users (question inline, may be fine).
- `ModifyPlan`'s hand-rolled replacement logic mis-plans the documented interpolated usage; `stringplanmodifier.RequiresReplace()` on the attribute handles it.
- `Update` sends the owner's `wif_project` for attached registrations, the same problem the settings resource had to fix for `wif_pool_name`.
- Making the settings WIF attributes Optional+Computed makes them impossible to unset, and the attached-registration guard only holds on the first write.
- The new pre-delete GET in the settings resource turns an already-deleted registration into a destroy failure.
- Import leaves `existing_wif_pool_id` null, so the next plan replaces a live registration.

Also worth noting: both tests that exercise attaching are skipped, so four attributes are changing their Required/Optional/Computed shape on a shared resource with no end-to-end coverage of the feature motivating the change. And this depends on an unreleased gofalcon pseudo-version.

internal/cloud_google_registration/cloud_google_registration.go
ffalor: this allows the requiresreplace logic in ModifyPlan to be deleted. A replace happens on the whole resource, it id deleted and recreated, so you dont need to mark multiple attributes as needs replace.

internal/cloud_google_registration/cloud_google_registration_settings.go
mzastavnyi-cs: Good catch! Dropped Computed from both attributes and fixed wrap() to null them out when attached

internal/cloud_google_registration/cloud_google_registration.go
mzastavnyi-cs: Added stringplanmodifier.RequiresReplace() plus UseStateForUnknown(), needed once wrap() populates this from the API on attach

internal/cloud_google_registration/cloud_google_registration.go
mzastavnyi-cs: Deleted

internal/cloud_google_registration/cloud_google_registration.go
mzastavnyi-cs: wrap() now sets existing_wif_pool_id from WifProperties.PoolID when WifPoolRegistrationID != "", null otherwise.

internal/cloud_google_registration/cloud_google_registration.go
mzastavnyi-cs: You're right, omitempty doesn't help. Guarded the assignment on !plan.ExistingWifPoolID.IsNull()

internal/cloud_google_registration/cloud_google_registration.go
mzastavnyi-cs: Gated both assignments in Update() behind plan.ExistingWifPoolID.IsNull()

internal/cloud_google_registration/cloud_google_registration.go
mzastavnyi-cs: Fixed

internal/cloud_google_registration/cloud_google_registration_settings.go
mzastavnyi-cs: Reverted Computed on both (back to Optional only), and fixed wrap() to null them out whenever WifPoolRegistrationID != "".

ashoksankarh-cs (approved):

alex-grant-cs (approved):

cs-wayne-hsieh (approved):

